PR #19, opened 2014-12-02
Can not use 'Save and Build'

NoahPeeters:
A Workspace with a new file (never saved before) could not be 'saved and build' in one step because the new file does not have a path. But when I start Atom and open a project there was always a new file.


noseglid:
I've implemented this feature in a5852d8. Closing this.

PR #32, opened 2015-01-10
Remember panel height when creating it

### LINKED ISSUE #31: Minimized Height option is ignored
When I change Minimized Height in settings, it does not apply unless I have checked "Keep Visible". Also after restarting Atom the height goes back to default but the value in settings remains the same.

Atom: 0.166.0
OS: Ubuntu 14.04

sryze:
Fixes #31

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -17,6 +17,7 @@ module.exports = (function() {
     this.titleLoopIndex = 0;
     this.a2h = new Convert();
     this.monocle = false;
+    this.attach();
 
     atom.config.observe('build.keepVisible', this.visibleFromConfig.bind(this));
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
@@ -46,6 +47,7 @@ module.exports = (function() {
 
   BuildView.prototype.attach = function() {
     this.panel = atom.workspace.addBottomPanel({ item: this });
+    this.height = this.output.offset().top + this.output.height();
   };
 
   BuildView.prototype.detach = function(force) {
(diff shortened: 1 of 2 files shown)


noseglid:
I didn't get around to give this a whirl yet, but just looking at the code it seems as the build panel will attach regardless of the `keepVisible` configuration.

I'll check this out in more depth tomorrow. Thanks for your contribution!

sryze:
It will attach, yeah, but then it'll also detach once `keepVisible` is read from the config.

It seems like there might be a problem with double attach though (if `keepVisible` is `true`). Should I check if the panel has been already created in attach?

lib/build-view.js
```diff
@@ -17,6 +17,7 @@ module.exports = (function() {
     this.titleLoopIndex = 0;
     this.a2h = new Convert();
     this.monocle = false;
+    this.attach();
```
noseglid: Is you acutely pointed out, when we call `atom.config.observe` with the keepVisible argument, `visibleFromConfig` will be called with the current choice the user have. This attach is not necessary then, because it will be either attach or detach att that point.

noseglid:
The problem with multiple attach, if it is that the `.bottom.tool-panel.panel-bottom` element is left (which gives an increasing black bar at the bottom). I found this while testing your fixed and corrected it in 8b182af1c5e33dfc805ede264578e8c8f6289352. (by always destroying it before creating a new one).

lib/build-view.js
sryze: But then if we remove this attach `this.height` will be uninitialized when `minimizedHeight` gets loaded from the config. 

The whole point was that we need to get the height of the panel while it still exists beacuse otherwise the height will be `0`. And since the panel might not have been ever made visible prior to reading `minimizedHeight` we need to explicitly create it somewhere.

Changing the order of `observe` calls might work too I guess, but this seemed more reliable and I wasn't sure if they are synchronous or not.

noseglid: Ah. I see.

You are correct. I think this is the best way too, and it works well. Thanks!

noseglid:
Thanks! I will include this in next release (v0.23.0).

PR #33, opened 2015-01-11
Auto build on save

### LINKED ISSUE #26: Adding option to build on save
Would be very useful to add a option to automatically build on save

sryze:
Adds new option "Auto Build On Save" that automatically triggers a build when a file gets saved.

Closes #26


sryze:
Would it be too bad to run the build if the user opens a totally unrelated file in the same Atom instance and saves it?

noseglid:
I think that's fine. How would we know if a file is unrelated anyway? It all depends on the build command, and the user can do virtually anything if specifying the build command with `.atom-build.json`.

This looks great. Thanks so much for this!

noseglid:
I changed the name of the variable to `buildOnSave` to be symmetrical with `saveOnBuild` in 4ab71c03228fa3d354084a037378071001b52a48.

PR #42, opened 2015-02-26
Add support for building Rust projects with Cargo

portify:
Use `cargo build` to build projects if `Cargo.toml` exists in the root


noseglid:
Thanks!

PR #43, opened 2015-02-28
Added setting to allow panel to remain open once process is killed

boxama:
This is a bit of a problem that I found when using this package. Since the process that I open doesn't seem to spew out the outputs I want it to until it closes (not the packages's fault), I never get enough time to inspect it.
I managed to edit the package and add a setting for this, I would appreciate it if this setting was added to the package to help others who encountered the same problem.
I see you're currently changing things around so I couldn't properly test with the current file. However, on the latest release these changes worked perfectly.

Edit: Just noticed a spelling mistake in "Visisble", should probably fix that too

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -45,6 +45,12 @@ module.exports = {
       default: 0.15,
       minimum: 0.1,
       maximum: 0.9
+    },
+    closeOnFinish: {
+      title: 'Close panel on finish',
+      description: 'Close build panel when the process is killed',
+      type: 'boolean',
+      default: true
     }
   },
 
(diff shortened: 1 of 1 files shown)


noseglid:
If you enable `keepVisible` the build window will remain open. That way you can inspect the output even if it succeeds.

If you close it with `keepVisible` enabled, it will be visible again, and remain visible until you close it. Isn't this what you want to achieve?

noseglid:
(I fixed the "Visisble" typo in 1984ddc, thanks for noticing this!)

boxama:
The problem with keepVisible is that whenever I start up Atom the panel is there, even though I might not be using it.

noseglid:
So if you're not using it just close it out? If you're not using it it wont come back until you trigger a build anyway (or toggle the keepVisible option).

It seems like what this PR is trying to accomplish is really close to what already exists, but just a different flavor. I'm not sure I want another configuration option for this minor difference.

PR #54, opened 2015-04-08
add support for scons

carneeki:
add support for scons

### DIFF
--- a/lib/tools/index.js
+++ b/lib/tools/index.js
@@ -12,6 +12,7 @@ module.exports = [
   require('./Grunt'),
   require('./gulp'),
   require('./make'),
+  require('./scons'),
 
   /* specific (builds only one thing) */
   require('./Elixir'),
--- a/lib/tools/scons.js
+++ b/lib/tools/scons.js
@@ -1,0 +1,14 @@
+'use strict';
+
+var fs = require('fs');
+
+module.exports.isEligable = function (path) {
+  return fs.existsSync(path + '/SConstruct');
+};
+
+module.exports.settings = function (path) {
+  return {
+    exec: 'scons',
+    sh: false
+  };
+};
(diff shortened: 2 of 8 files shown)


noseglid:
Cool. Would you mind adding a test for SCons as well?

carneeki:
I'm not 100% familiar with the test environment and I got about halfway through before I thought I should ask for advice.

I've written a few SConstruct files based on the Makefiles provided, and I started modifying ```/spec/build-spec.js``` but I'm not sure I'm on the right track. Is build-spec.js automatically generated?

Also... Ideally, I need to run scons with a few extra flags compared to normal:
```scons -sc distclean -f SConstruct.escape```

This is because scons generates a checksum file, and ideally it would be nice to clean that up after each test.

What are your thoughts?

noseglid:
It looks good! `build-spec.js` is not automatically generated, so what you're doing is the right thing.

It appears as the `SConstruct.escape` file (and the defined variable) is never used. Should it be removed?

Each test is run in a new clean directory which is removed after the test is done. Are the checksum files kept in a global state or would they be removed along with the directory?

carneeki:
Cool thanks for the advice :)

I hadn't got around to using ``` SConstruct.escape``` yet. I'll keep
working on it.

The checksum files are OK to let the framework remove. I just thought I had
to tidy up after myself, but if the framework does it, that's even better.

Thanks Alexander!

On 9 April 2015 at 15:57, Alexander Olsson <notifications@github.com> wrote:

> It looks good! build-spec.js is not automatically generated, so what
> you're doing is the right thing.
>
> It appears as the SConstruct.escape file (and the defined variable) is
> never used. Should it be removed?
>
> Each test is run in a new clean directory which is removed after the test
> is done. Are the checksum files kept in a global state or would they be
> removed along with the directory?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/noseglid/atom-build/pull/54#issuecomment-91123171>.
>



-- 
Adam "carneeki" Carmichael
p: +61 415 37 1990
w: http://bigneek.com
e: carneeki@carneeki.net
i: 2207644

noseglid:
Any updates on this?

noseglid:
Closing this out. Please re-open if you wish to finish this PR.

PR #57, opened 2015-04-16
Cycling through match errors

5n8ke:
Instead of only allowing to jump to the first location found by the given errorMatch regular expression, you can cycle through every error with build:error-match. At the end it loops around to the beginning.
build:error-match-first (cmd-alt-h) jumps directly back to the beginning.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -61,6 +61,7 @@ module.exports = {
     this.stderr = new Buffer(0);
     atom.commands.add('atom-workspace', 'build:trigger', this.build.bind(this));
     atom.commands.add('atom-workspace', 'build:error-match', this.errorMatch.bind(this));
+    atom.commands.add('atom-workspace', 'build:error-match-first', this.errorMatchFirst.bind(this));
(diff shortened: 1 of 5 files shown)


lib/build.js
```diff
@@ -135,27 +137,63 @@ module.exports = {
     return value;
   },
 
+  errorParse: function() {
+    this.match = [];
+    var regex = XRegExp(this.cmd.errorMatch);
```
5n8ke: Depending on the regex used the multiline flag possibly needs to be added. But with quick and dirty regex for g++ it is not needed.

noseglid:
This looks great. Would you mind adding some tests to this?

5n8ke:
I hope these tests are fine, it's the first time creating some for me. It tests cycling and wrap around and jumping to the first error match.

I don't know why travis failed. In my own repo the build and test are without error.

noseglid:
I just re-started the build. Seems as there was an outage in the travis service.

noseglid:
Thanks you very much for this contribution. It will make an excellent addition to atom-build!
It will be included in the next version (`v0.29.0`).

PR #60, opened 2015-04-26
Add build:toggle-view command

sryze:
This command allows you to show or hide the build view. The default shortcut is `cmd-alt-b cmd-alt-v` (Mac) or `ctrl-alt-b ctrl-alt-v` (others).

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -22,6 +22,8 @@ module.exports = (function() {
     atom.config.observe('build.panelVisibility', this.visibleFromConfig.bind(this));
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
     atom.config.observe('build.minimizedHeight', this.heightFromConfig.bind(this));
+
+    atom.commands.add('atom-workspace', 'build:toggle-view', this.toggle.bind(this));
   }
 
   cscompatability.extends(BuildView, View);
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -23,7 +23,7 @@ module.exports = (function() {
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
     atom.config.observe('build.minimizedHeight', this.heightFromConfig.bind(this));
 
-    atom.commands.add('atom-workspace', 'build:toggle-view', this.toggle.bind(this));
+    atom.commands.add('atom-workspace', 'build:toggle-panel', this.toggle.bind(this));
   }
 
   cscompatability.extends(BuildView, View);
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -68,6 +68,10 @@ module.exports = (function() {
     }
   };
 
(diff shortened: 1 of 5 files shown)


sryze:
I guess the shortcut should be changes, currently it makes build:trigger wait for the second key press for a little while before building.

noseglid:
I like this! 

It's smart to use `*-alt-v` so it doesnt freeze `*-alt-b` for a second or so.

I've recently made some changes on how visibility is set (due to #62). Is this compatible with those changes??

lib/build-view.js
```diff
@@ -95,6 +99,14 @@ module.exports = (function() {
     this.detach(true);
   };
 
+  BuildView.prototype.toggle = function(event, element) {
+    if (this.isAttached()) {
+      this.detach(true);
+    } else {
+      this.attach();
```
noseglid: force flag has been added to attach as well in commit cc66db0 this must be passed here as well.

Also, I think this can be written neater by

    this.isAttached() ? this.detach(true) : this.attach(true);

lib/build.js
```diff
@@ -60,6 +60,7 @@ module.exports = {
     this.stdout = new Buffer(0);
     this.stderr = new Buffer(0);
     atom.commands.add('atom-workspace', 'build:trigger', this.build.bind(this));
+    atom.commands.add('atom-workspace', 'build:toggle-view', this.toggleView.bind(this));
```
noseglid: Maybe this command should be added in `build-view.js` instead since it's purely a presentation thing?

This would also keep us from having to implement the simple pass-through function `toggleView` in this class.

noseglid:
I tested it, and it almost works. `attach` needs the force flag as well.
Have a look at my comments. If you agree, fix them - otherwise argue why you shouldn't :)

Thanks for the effort in implementing this!

lib/build-view.js
sryze: I guess that one looks better, yeah. Though I've mostly seen the ternary operator used when something is actually being returned (i.e. as part of an expression).

lib/build.js
sryze: Oh, didn't know it's ok to register commands in `build-view.js` (I guess that's because there were none).

But it feels kind of wrong to register commands in the build view. I mean, it looks like we can create multiple instances of BuildView but Atom commands seem kind of like global things, wouldn't that create problems?

sryze: Never mind, it seems that registering the same command multiple times is not an issue according to the docs.

sryze:
Should I also rename this to `toggle-window`? I just noticed that it's called a window everywhere instead of a view (kind of confusing given the class name).

noseglid:
Or build panel (the thing we show is internally called a `panel` in atom).

sryze:
OK, done.

noseglid:
Excellent. Great job! I will include it in the next version `v0.31.0`.

PR #64, opened 2015-05-01
Add timeout to dismiss build view

detj:
#### What is changed
Gives control of the build view dismiss time interval back to users. Reads optional key `buildViewTimeout` from `atom-build.json`. Default is 1 second.

Let me know if you would like to name the key something else and I'll follow suit. Any other comments/suggestions are welcome.

#### Reason
I was building a Go file one at a time using `go run` and wanted to analyse the output in the build view without switching windows. But the build view was closing after 1 second making it difficult for me to look at the output of the program.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ exactly what to execute. Create a file named `.atom-build.json` in your project 
         "VARIABLE2": "VALUE2",
         ...
       },
-      "errorMatch": "^regexp$"
+      "errorMatch": "^regexp$",
+      "buildViewTimeout": 1000
     }
 
 Note that if `sh` is false `cmd` must only be the executable - no arguments here. If the
@@ -65,6 +66,7 @@ systems).
   * `cwd` - **[optional]** The working directory for the command. E.g. what `.` resolves to.
   * `env` - **[optional]** An array of environment variables and their values to set
   * `errorMatch` - **[optional]** A regular expression to match output to a file, row and col. See [Error matching](#error-match) for details.
+  * `buildViewTimeout` - **[optional]** Interval in milliseconds to dismiss build view
 
 ### Replacements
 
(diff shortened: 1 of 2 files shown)


sryze:
You can also use the `toggle-panel` command or `ctrl-alt-v` with latest commits (it's not released yet). I'm not telling that this is not needed though, having the timeout configurable may be still useful.

detj:
Good to know that. I'm not on the latest release.

README.md
```diff
@@ -48,7 +48,8 @@ exactly what to execute. Create a file named `.atom-build.json` in your project
         "VARIABLE2": "VALUE2",
         ...
       },
-      "errorMatch": "^regexp$"
+      "errorMatch": "^regexp$",
+      "buildViewTimeout": 1000
```
noseglid: I don't think you'd want to configure this differently for different builds. I think this should be a configuration which is global instead.

README.md
```diff
@@ -65,6 +66,7 @@ systems).
   * `cwd` - **[optional]** The working directory for the command. E.g. what `.` resolves to.
   * `env` - **[optional]** An array of environment variables and their values to set
   * `errorMatch` - **[optional]** A regular expression to match output to a file, row and col. See [Error matching](#error-match) for details.
+  * `buildViewTimeout` - **[optional]** Interval in milliseconds to dismiss build view
```
noseglid: You should add some thing like " ... upon successful build.'. This probably also needs some clarifications on how it works in conjunction with the configuration variable `panelVisibility`, which can be `Toggle`, `Keep Visible` or `Show on Error`.

detj: Okay. Let me update the PR

noseglid:
Any updates on this?

noseglid:
Closing this out. Please re-open if you wish to finish this PR.

PR #69, opened 2015-05-09
Add uniform syntax for environment variables

sryze:
Replace occurrences of `$XYZ` with corresponding environment variables. This basically allows you to use the same environment variable syntax in `cmd` on Windows as you would do on Linux (and probably OS X too) with `sh: true`, and thus allows you have one atom-build.json file for both platforms.


noseglid:
This looks good. I would love to see a test for this as well.

sryze:
Sorry for such a long delay. I extended the replace test a little bit to check if variables from process' environment can pass through, would this be enough?

noseglid:
Thanks! I'll have this included in the next release (v0.35.0).

PR #71, opened 2015-05-12
Added changes to accept absolute error file path

mirhec:
@see Readme.md

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -175,8 +175,9 @@ module.exports = {
 
   errorMatchGoTo: function () {
     var file = this.replace(this.cmd.cwd) + path.sep + this.match[0].file;
-    if(!this.match[0].file)
+    if (!this.match[0].file) {
       file = this.match[0].absFile;
+    }
     atom.workspace.open(file, {
       initialLine: this.match[0].line ? this.match[0].line - 1 : 0, /* Because atom is zero-based */
       initialColumn: this.match[0].col ? this.match[0].col - 1 : 0 /* Because atom is zero-based */
--- a/lib/build.js
+++ b/lib/build.js
@@ -176,8 +176,7 @@ module.exports = {
   errorMatchGoTo: function () {
     var file = this.replace(this.cmd.cwd) + path.sep + this.match[0].file;
     if(!this.match[0].file)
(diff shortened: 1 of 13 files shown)


noseglid:
I'm not really sure what you want to achieve here. It seems like a really convoluted way to just specify that the matched file is an absolute path. A simple boolean flag, telling if the path is absolute or not seems like it would be enough.

Besides. Why is your build tool outputting absolute paths when you're clearly working in a specific directory? What build tool is this?

mirhec:
Sorry for the few informations I provided so far.

The build system I was using is the Visual Studio 6 Compiler (msdev) on Windows. Unfortunately it always prints out an absolute path instead of a relative one. Here is an example output:

```
Executing: msdev  V7RtSpcX.dsp /MAKE V7RtsPcX - Win32 Release /BUILD
--------------------Konfiguration: V7RtSPCX - Win32 Release--------------------
Kompilierung l�uft...
pccontro.cpp
K:\VISION7\STEU\pccontro.cpp(3243) : error C2065: 'bug' : nichtdeklarierter Bezeichner
Fehler beim Ausf�hren von cl.exe.
Browse-Informationsdatei wird erstellt...

V7RtSPCX.dll - 1 Fehler, 0 Warnung(en)
```

I noticed that the changes I made to find the `.atom-build.json` file was not necessary, so I removed it.

lib/build.js
```diff
@@ -124,7 +124,8 @@ module.exports = {
       args: [],
       cwd: path,
       sh: true,
-      errorMatch: ''
+      errorMatch: '',
+      errorMatchPathAbsolute: false
```
noseglid: To avoid having more configuration options (which tend to bloat the README and files) I want to suggest having a different name for the matched group in the regex instead. It could be called `absFile`. When error matching, `atom-build` should check if `file` is matched, in that case open `cwd + file`, otherwise of `absFile` is set, open `absFile`.

mirhec:
I changed the code like you suggested.

noseglid:
Looks great. Fix the code style hints and I'd be happy to merge it.

noseglid:
Thank you for this!

Issue #37, opened 2015-01-24
Add support for multiple build commands


sryze:
It would be great if we could define multiple build commands in `.atom-build.json` instead of just one. This would allow e.g. creating separate bindings for different commands, like `make` and `make test`.

I've seen something like this in [SciTE](http://www.scintilla.org/SciTE.html) where you can assign numbers to different commands and then use `Ctrl+N` to trigger command `N`.

noseglid:
Cool idea. I really like it.

Maybe it should be possible to extend the `.atom-build.json` file with a keymap where the value is the same structure as the `.atom-build.json` is defined to be today. That way we could have different environments and arguments and all the goodies available in the plugin today.

AngelEzquerra:
This could also be used to create a "run" command in addition to a "build" command.
Ideally, you should be able to also define a separate key combination for each such command.

austinbirch:
I’ve just started using Atom, found this package, and then needed the ability to specify multiple builds. I hacked something together, but not sure whether this is the sort of the implementation you would like?

Here’s my changes: https://github.com/austinbirch/atom-build/compare/multiple-build-commands

If this is the sort of thing you are looking for, I’d be happy to add tests, make any changes, and submit a pull request.

noseglid:
@austinbirch I really like what you're proposing!

Some notes which I may not have thought all the way through:
  * You're reading the `.atom-build.json` when the package is activated (which for most users is when Atom starts). Would it be good to re-read that file,

austinbirch:
@noseglid Thanks for the feedback!

First reaction to your notes (also not thought all the way through!):

> Maybe we should have another file name than .atom-build.json for the named builds.

I think I’d prefer having one file that defines all builds, though I’m not sure what the best implementation here is. In my changes I’ve allowed the `.atom-build.json` to define an array of builds, though it feels a little odd to allow both a single object and an array in that config file.

The problem I’m seeing with using an array is that I’m not sure what the `build:trigger`, `build:confirm`, `build:no-confirm` actions should do in this case (in my changes, I’ve just allowed a user to specify a "default" that is used for those commands. This is not a great solution, because you could theoretically define more than one build as the default).

Not sure how open you would be to changing the format of the config file, but perhaps it could be something like below.

```
{
   "default": "name1",
   "otherBuildsSettingHere": "value",
   ...,
   "builds": {
       "name1": {...build config...},
       "name2": {...build config...}
   }
}
```

[...]

PR #77, opened 2015-05-18
Build targets

austinbirch:
Basic functionality to solve #37.

**Notes:**

- The user can deselect a target (so that the `build:trigger` command will run the default, unnamed, build)  by selecting it again in the `TargetsView`. Not sure what the atom UI paradigm for toggling in a list view is, so we might want to change this. Maybe we could have a "Default" option always available, or there might be a UI convention for this already?
- Need to implement returning targets from other tools, currently only returns targets from the `atom-build` tool.
- Needs some tests. Once we have settled on the functionality and implementation, I’m happy to write  the tests before merging.


noseglid:
Cool! Thanks a lot. I think we should use this as a base for the rest (more tests, README, code style and eventual other things). I've pushed this to `dev/build-targets`.

I suggest we change the key-binding to `cmd-alt-t` (OS X) or `ctrl-alt-t` (Linux/Windows), so it has the same meta keys as running the build target.

I really love `mark-active` part of SelectListView. Good job on finding that. Looks great.

noseglid:
This has been merged to `noseglid/atom-build@dev/multiple-targets`. We should continue work there (I've pushed your commits, so you will be attributed for the changes you've made so far, and any further changes you'll make).

PR #88, opened 2015-05-30
Panel Orientation

Zilarrezko:
Alright, so I'm looking through the files and I can see most of what I need to change to make this happen I believe is in build.js and build-view.js. So I looked through, trying to see what's what. Trying to insert what I think is necessary.

I'm sure I missed some things, and I'd like so mention what I know to be a bug or something.

First thing... I made a spelling mistake, I fixed it... But it seems to keep this around:
something_wrong

Next, I was looking around, because  when I built when in left or right orientation, it took up a very large portion of the window. That's useless because I have the "height" at a very low percentage. I thought if I changed the css output to width and/or changed this.output.height to this.output.width it would change it but I'm ignorant in this field. It obviously didn't work out too well, now I can't get it big enough.

I did however succeed in getting the panel to orientate where I wanted. So that's cool. Obviously though, I'd like to get these things fixed.

Also, I didn't know what to rename height to. I was thinking length, but I thought that could be misleading as well. But I'm not sure.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -60,14 +60,17 @@ module.exports = (function() {
 
     if ('Top' === atom.config.get('build.panelOrientation')) {
         this.panel = atom.workspace.addTopPanel({ item: this });
+        this.height = this.output.offset().top + this.output.width();
     } else if ('Bottom' === atom.config.get('build.panelOrientation')) {
         this.panel = atom.workspace.addBottomPanel({ item: this });
+        this.height = this.output.offset().top + this.output.width();
     } else if ('Left' === atom.config.get('build.panelOrientation')) {
         this.panel = atom.workspace.addLeftPanel({ item: this });
+        this.height = this.output.offset().top + this.output.width();
     } else if ('Right' === atom.config.get('build.panelOrientation')) {
         this.panel = atom.workspace.addRightPanel({ item: this });
+        this.height = this.output.offset().top + this.output.width();
     }
-    this.height = this.output.offset().top + this.output.height();
   };
 
   BuildView.prototype.detach = function(force) {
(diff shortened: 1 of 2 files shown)


lib/build-view.js
```diff
@@ -56,8 +57,20 @@ module.exports = (function() {
     if (this.panel) {
       this.panel.destroy();
     }
-    this.panel = atom.workspace.addBottomPanel({ item: this });
-    this.height = this.output.offset().top + this.output.height();
+
+    if ('Top' === atom.config.get('build.panelOrientation')) {
```
noseglid: I think a switch-case would be more appropriate here.

Also, I think the default value should be bottom - to keep backward compatability. This is also likely the cause of all the broken tests.

lib/build.js
```diff
@@ -21,26 +21,34 @@ module.exports = {
       enum: [ 'Toggle', 'Keep Visible', 'Show on Error' ],
       order: 1
     },
+    panelOrientation: {
+      title: 'Panel Orientation',
+      description: 'Set where the build panel should be.',
+      type: 'string',
+      default: 'Bottom',
+      enum: [ 'Top', 'Bottom', 'Left', 'Right' ],
```
noseglid: I don't think we should support Left or Right. I can not see any useful scenario where you would like to add it to the left or right of the screen. Since text is left-to-right there would **a lot** of line breaks.

This option should also be added as order 7 to reduce the diff - it is also less likely that people will change this.

Zilarrezko: I understand. My original thought it was to have an option for the people that liked using the split function for their code between left and right. However, it looks like even if they had a split it would create a third window in the split, and you're right. To read the error _and_ look at the now squished windowed code to fix it an error, is laughably awkward.

I'll make try and make the changes to get it to orientate to the top/bottom if you still want it. Though that task is so trivial you could just do it yourself and save yourself the pull. Sorry for the time waster man.

noseglid: I'd love to see an updated PR!

I also want to see a simple test which tests that setting the orientation actually attaches the panel at different locations. Take other tests as inspiration. It is probably easiest to check the classes on the build panel (I'm guessing Atom adds different classes depending if it is a `TopPanel` or a `BottomPanel`).

jhasse: Actually supporting Left or Right makes a lot of sense when using an ultra-wide monitor, doesn't it?

Zilarrezko: It does for my setup at least.

setup

noseglid: I wouldn't mind support for left or right as well. Include it if you like!

I'm not gonna budge on the test though :tent:

noseglid:
Closing this out due to inactivity. Please re-open if you continue on this.

PR #90, opened 2015-06-04
Error auto parse

5n8ke:
Added config to automatically parse and scroll to the first matched error after a build fails.


noseglid:
This looks good! Could you also add a test for this?

5n8ke:
I think the tests should be fine that way.
I am not pleased with the fact, that the config is not part of the errorMatcher module, but I don't think there is an easy way to move it out of the main package.

spec/build-error-match-spec.js
```diff
@@ -267,7 +267,33 @@ describe('Build', function() {
...
+
+      runs(function() {
+        var editor = atom.workspace.getActiveTextEditor();
+        var bufferPosition = editor.getCursorBufferPosition();
+        expect(editor.getTitle()).toEqual('.atom-build.json');
+        expect(bufferPosition.row).toEqual(2);
+        expect(bufferPosition.column).toEqual(7);
+      });
 
+      atom.config.set('build.scrollOnError', true);
```
noseglid: This should probably not be here.
You should also set it to the default value in the `beforeEach` (which is false in this case).

noseglid:
Excellent. Just a minor review comment then we're good to merge!

5n8ke:
Fixed my copy and paste error and added the default config in the beforeEach.

spec/build-error-match-spec.js
```diff
@@ -267,7 +268,33 @@ describe('Build', function() {
...
+
+      runs(function() {
+        var editor = atom.workspace.getActiveTextEditor();
+        var bufferPosition = editor.getCursorBufferPosition();
+        expect(editor.getTitle()).toEqual('.atom-build.json');
+        expect(bufferPosition.row).toEqual(2);
+        expect(bufferPosition.column).toEqual(7);
+      });
 
+      atom.config.set('build.scrollOnError', false);
```
noseglid: This line shouldn't be here at all.
Since waitsFor and runs are async, this line will reset the configuration to false before your test even runs.

After putting the configuration setting in `beforeEach`, we've made sure each test always start from the same fixture. There's no need to "reset" the configuration here.

5n8ke:
Ah, now I understand. That is the problem of working at two in the night :smile:

noseglid:
Splendied! Thank you so much for this!

PR #93, opened 2015-06-07
Fix bug that prevents error matching on successfull build command

5n8ke:
In it's previous place these two lines only allowed error matching if the build command returned a nonzero value.

This prevented cycling through warnings that didn't make the build exit with an error.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -300,14 +300,16 @@ module.exports = {
 
         this.child.on('close', function(exitCode) {
           this.buildView.buildFinished(0 === exitCode);
+
+          var t = this.targets[this.activeTarget];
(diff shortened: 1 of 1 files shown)


5n8ke:
The Travis build worked in my own repository :/
[Build Status](https://travis-ci.org/5n8ke/atom-build)

But in my local repository it failed one time before I applied my patch on the same line (build.js:234) too.

noseglid:
Intermittent issues like that often means race conditions.
I checked out your PR and ran the tests, it works *sometimes*. It appears that a lot of tests clean up the Atom environment simultaneously as the `close` callback on `this.child` is being called.

Not sure what the best way to handle being in a callback while Atom is being torn down is. Suggestion?

noseglid:
Closing due to no response.

PR #96, opened 2015-06-10
Use fs.realpath to detect absolute paths, fix #94

### LINKED ISSUE #94: Should detect absolute path
If a file path starts with `/` it should be treated as an absolute one. At the moment I'm getting:

```
Matched file does not exist: /home/jhasse/git/gtoc8//usr/include/boost/test/floating_point_comparison.hpp
```

jhasse:
I'm really no JavaScript developer and haven't done any Atom or Node related stuff, so you better test this ;)

### DIFF
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -23,12 +23,21 @@ module.exports = (function () {
   util.inherits(ErrorMatcher, EventEmitter);
 
   ErrorMatcher.prototype._goto = function () {
-    var file = this.cwd + path.sep + this.currentMatch[0].file;
-    if (!this.currentMatch[0].file) {
+    var file = this.currentMatch[0].file;
+    if (file) {
+      var oldDir = process.cwd();
+      process.chdir(this.cwd);
+      try {
+        file = fs.realpathSync(file);
+      } catch (e) {
+        file = this.cwd + path.sep + file;
+      }
+      process.chdir(oldDir);
+    } else {
       file = this.currentMatch[0].absFile;
-    }
-    if (!file) {
-      return this.emit('error', 'Did not match any file. Don\'t know what to open.');
+      if (!file) {
+        return this.emit('error', 'Did not match any file. Don\'t know what to open.');
+      }
     }
 
     var row = this.currentMatch[0].line ? this.currentMatch[0].line - 1 : 0; /* Because atom is zero-based */


jhasse:
New commit uses realpathSync instead

jhasse:
Travis CI is great :) New commit adds space before catch.

jhasse:
Sorry for all the commits! I've got `npm test` working on my computer, but running the Atom tests doesn't, so I have to wait for the CI results.

This last one should work :)

lib/error-matcher.js
```diff
@@ -23,12 +23,21 @@ module.exports = (function () {
   util.inherits(ErrorMatcher, EventEmitter);
 
   ErrorMatcher.prototype._goto = function () {
-    var file = this.cwd + path.sep + this.currentMatch[0].file;
-    if (!this.currentMatch[0].file) {
+    var file = this.currentMatch[0].file;
+    if (file) {
+      var oldDir = process.cwd();
+      process.chdir(this.cwd);
```
noseglid: Instead of changing directory for the whole process, wouldn't it be enough to test if `fs.realpathSync(this.cwd + path.sep + this.currentMatch[0].file)` exists? Then you wouldn't have to change back cwd at the end of this function either.

jhasse: Yes, I did remove the path.sep path of the code but that caused the test to
fail. My suggestion would be to remove absFile from the spec, then the
whole match function can be simplified even further.

noseglid: I agree that we should look to deprecate `absFile` in regard to this change.

I still don't see why you need to `chdir`. Check out [these changes I pushed to a dev branch](https://github.com/noseglid/atom-build/commit/a452263491dee7e9d8b93ef04b370ddf8a188cf6). Wouldn't that do it?

jhasse: Yes, that looks good :) Didn't think of that

jhasse: Wait! What if there's a file with same name in process.cwd() and in
this.cwd?

noseglid: Well, it seems weird if `this.cwd` and `process.cwd()` differs AND the build tool outputs a file relative to `process.cwd()` when the working directory is set to `this.cwd`. Essentially, the executed build command knows nothing of `process.cwd()`.

jhasse: Think of a common file name like Makefile or main.cpp. It's unlikely, but
possible.

noseglid: Well, it would open the file which is relative to the working directory of the build tool - which probably is what we want, no?

jhasse: It would open the file relative to process.cwd() as that's what
realpathSync is using.

noseglid: Right, you are correct!
It would unfortunate, but not unlikely (as you said, `main.cpp` is a common file).
Is changing the process working directory the only way?

jhasse: I guess so, the only other way I can think of is re-implement the realpathSync.

Is changing the process cwd actually a bad idea? Is Atom multi-threaded?

noseglid: Well no. But I tend to refrain from doing stuff which change global behavior. If a user uses any other package which depends on current working directories, it's unlikely that package will work well with build.

Here's another solution. We could use [fs.isAbsolute](https://nodejs.org/api/path.html#path_path_isabsolute_path) to check if path is absolute. This takes into account all different platforms node is supported on . If the path is absolute then we just use `path` otherwise we use `this.cwd + path.sep + path`. What do you think?

jhasse: The global behavior isn't changed by my patch since I change the cwd back after the realpathSync call. But `fs.isAbsolute` actually looks like a good and clean solution, so to be safe let's get rid of `process.chdir` :). It also makes it clearer what's going on.

noseglid: Do you feel like doing the changes to `fs.isAbsolute`? Otherwise I'll just do it.

jhasse: I think it's better if you do it, otherwise I have to commit like a thousand times again and wait for Travis :D

noseglid:
Closing this in favor of 0f4327d (solves #94).

PR #97, opened 2015-06-14
Search all panes when opening a matched file

jhasse:
A simple change, which makes working with multiple panes and atom-build easier. Before this patch I would end up with multiple files opened in all panes.


noseglid:
Thanks for this! :+1:

Issue #109, opened 2015-06-26
package load time is high


braver:
atom-build takes about 170ms load time on my machine, which is a bit much. There is a lot of stuff going on before package activation that might be postponed. I was able to get [terminal-panel](https://github.com/thedaniel/terminal-panel/pull/67), which had similar activation times and a lot of requires, down to 5ms on activation by postponing everything until the package functionality is actually used. I think I might look into it if you don't mind, but also putting it on your radar just to be sure.

noseglid:
I'd be delighted! I've been looking at timecop myself and thought along those same lines! Your help would be much appreciated.

noseglid:
This improved some with the removal of bluebird (d3693f3). Still more to do though.

noseglid:
Fixed with #352

Issue #108, opened 2015-06-26
build panel shows on activate when "keep visible"


braver:
If you select the "Keep visible" mode, I would expect the panel to keep visible after a build is triggered and completed successfully. However, it currently has that side effect of the panel opening on activation of the package, which means it opens with every new Atom window, even though no build is triggered.

Atom 1.0, atom-build 0.38, OSX 10.10.4

noseglid:
Fixed by 938804b (#115)

PR #111, opened 2015-06-26
wip: load&activate performance

braver:
for #109 

Shifting some requires around yields some improvement, but also adds some duplication. The real win would be in not creating a BuildView on activate but on the first time a build is run. That would drop the activation time from ~ 150 to 15ms (load time around 40ms). Coincidentally that would also solve #108 where a build panel is shown even though no build has been run.
There is no need to run create any views until there is a need to do so, so that should be possible (with some work). That mirrors my experience with terminal-panel, where creating the view was the most expensive as well.

Edit: my tinkering is mostly inconsequential, some milliseconds here and there. Ideally BuildView should only be done the first time a build is run, that would save almost 150ms, of 75% of the current startup time for atom-build. It would take me a bit more time than I have right now to figure out how to do that, but perhaps it sets you on your way.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -18,7 +18,6 @@ module.exports = (function() {
     this.titleLoopIndex = 0;
     this.a2h = new Convert();
     this.monocle = false;
-    this.attach();
 
     atom.config.observe('build.panelVisibility', this.visibleFromConfig.bind(this));
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
@@ -94,10 +93,6 @@ module.exports = (function() {
         if (!this.title.hasClass('error')) {
           this.detach();
         }
-        break;
-
-      case 'Keep Visible':
-        this.attach();
         break;
     }
   };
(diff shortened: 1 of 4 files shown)


noseglid:
Are you planning on continuing on this, or did you merely want this ball to get rolling?

braver:
I kinda plan to continue this, but it depends on if it's going to be me hitting my head against a wall another night or if I'll actually get some real progress. At this point I'm a bit afraid it's going to be quite a large refactor.

braver:
@noseglid I'm sorry but I'm pulling out of this. I'll end up rewriting so much and it's just going to take me too much time even if I manage to do it without breaking the entire thing. I'll leave the issue open, but this PR is going nowhere.

noseglid:
@braver Quite alright. I've continued on some of your work and I think I might have something fairly useful. I managed to get both load times and activation times down to below 5ms (which excludes them from the TimeCop view).

Thanks for the effort anyway. It got me thinking more about it and you gave me an excellent starting point!

braver:
Great, glad to be of help!

PR #115, opened 2015-06-28
don't attach view on package activation

braver:
fix for #108 

I think the idea behind the option "Keep visibile" would be to keep the build output visible after a build, not to have the build output panel visible at all times. Package activation happens each time a new atom window is opened and it doesn't seem useful to me to have the build panel open right away (with no output).

I could be wrong though, and perhaps another option is needed. In any case, this PR changes the behaviour to keep the panel visible after a successful build, but not immediately after package activation.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -18,7 +18,7 @@ module.exports = (function() {
     this.titleLoopIndex = 0;
     this.a2h = new Convert();
     this.monocle = false;
-    this.attach();
+    // this.attach();
 
     atom.config.observe('build.panelVisibility', this.visibleFromConfig.bind(this));
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
@@ -95,10 +95,10 @@ module.exports = (function() {
           this.detach();
         }
         break;
-
-      case 'Keep Visible':
-        this.attach();
-        break;
+      //
+      // case 'Keep Visible':
+      //   this.attach();
+      //   break;
     }
   };
 
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -18,7 +18,6 @@ module.exports = (function() {
     this.titleLoopIndex = 0;
     this.a2h = new Convert();
     this.monocle = false;
-    // this.attach();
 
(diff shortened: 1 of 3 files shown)


spec/build-visible-spec.js
```diff
@@ -60,10 +60,7 @@ describe('Visible', function() {
 
     describe('when build panel is toggled and it is visible', function() {
       beforeEach(function () {
-        atom.config.set('build.panelVisibility', 'Keep Visible');
```
braver: In the specs the "keep visible" option is used as a way to make the panel visible. Instead, toggle-panel can be triggered.

spec/build-visible-spec.js
```diff
@@ -45,8 +45,8 @@ describe('Visible', function() {
       });
     });
 
-    it('should show build window', function() {
-      expect(workspaceElement.querySelector('.build')).toExist();
+    it('should not show build window', function() {
```
braver: The new expectation is to not have the panel visible on package activation.

lib/build-view.js
```diff
@@ -18,7 +18,6 @@ module.exports = (function() {
     this.titleLoopIndex = 0;
     this.a2h = new Convert();
     this.monocle = false;
-    this.attach();
```
braver: So it won't attach the view on activation.

lib/build-view.js
```diff
@@ -95,10 +94,6 @@ module.exports = (function() {
           this.detach();
         }
         break;
-
```
braver: It's not really necessary to open the panel when the config changes. Also, config subscriptions are triggered on package activation, so anything you do here also happens every time an Atom window is opened (not just when the value changes). So it's better to not do too much here.

braver:
As a bonus, this shaves off almost 50ms in package activation.

screen shot 2015-06-28 at 12 32 49

noseglid:
This is really cool. I do realize that we are changing the behavior with this change, but I agree with you that it makes more sense to show the panel on build, and then simply keep it. There's really no real need to have an empty build show up on activation!

Thanks a lot for this!

braver:
:+1: awesome!

PR #117, opened 2015-06-30
Atom Linter support for build errors/warnings

debuuu:
I added support for pushing errors/warnings to Atom's Linter package, currently only for the Rust extension. I created a _linthelper.js_ class under tools, which can handle parsing the regex and building an array of Linter-complaint objects, and pushing them to that package.

I added a _linter_ property to your extension settings (currently only in the _Cargo.js_ one), which is then queries in the _build.js_ whenever a build is requested, and if it exists it updates the lint. 

I'm not sure if it covers all bases, but anyway, let me know if you think this is something that could be merged!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Atom Build package
-[![Build Travis](https://travis-ci.org/noseglid/atom-build.svg?branch=master)](https://travis-ci.org/noseglid/atom-build)
+[![Build Travis](https://travis-ci.org/noseglid/atom-build.svg?branch=master)](https://travis-ci.org/debuuu/atom-build)
 [![Gitter chat](https://badges.gitter.im/noseglid/atom-build.svg)](https://gitter.im/noseglid/atom-build)
 
 Automatically build your project inside your new favorite editor, Atom.
--- a/lib/build.js
+++ b/lib/build.js
@@ -310,12 +310,12 @@ module.exports = {
           this.buildView.append((this.cmd.sh ? 'Unable to execute with sh: ' : 'Unable to execute: ') + this.cmd.exec);
           this.buildView.append(/\s/.test(this.cmd.exec) ? '`cmd` cannot contain space. Use `args` for arguments.' : '');
         }.bind(this));
-        
+
         this.child.on('close', function(exitCode) {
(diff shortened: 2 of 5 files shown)


noseglid:
There were some [linting errors](https://travis-ci.org/noseglid/atom-build/builds/69046359) in your code. Please fix these.

debuuu:
Ah! Ok, I'll look into those today if I get time. I don't work with Javascript too often, so I'm not sure of what workflows exist for code quality checking - sorry about that.

README.md
```diff
@@ -1,5 +1,5 @@
 # Atom Build package
-[![Build Travis](https://travis-ci.org/noseglid/atom-build.svg?branch=master)](https://travis-ci.org/noseglid/atom-build)
+[![Build Travis](https://travis-ci.org/debuuu/atom-build.svg?branch=master)](https://travis-ci.org/debuuu/atom-build)
```
noseglid: You should not change this.

lib/build.js
```diff
@@ -61,6 +61,13 @@ module.exports = {
       default: true,
       order: 5
     },
+    showBuildNotifications: {
```
noseglid: Why do you want to add notifications? The build panel should be notification enough, no?

lib/build.js
```diff
@@ -311,17 +321,32 @@ module.exports = {
         this.child.on('close', function(exitCode) {
           this.buildView.buildFinished(0 === exitCode);
           if (0 === exitCode) {
+            // if the command has a linter, clear lints for successful builds.
+            if (this.cmd.linter !== undefined) {
```
noseglid: It seems very wrong to have one linter per command. I thinks it feels more logical if there is one linter which is used for whatever command is run, and cleared between runs.

This would also help remove the Rust specific code in `linthelper.js` and the `Cargo` build tool

colin-kiegel:
I wonder whether this could lead to duplicate lint-messages when combined with other linter-tools.

But I like the general idea - it should probably be optional if duplicate lints could be an issue.

noseglid:
Severely outdated. Please open a new PR if you intend to continue on this.

PR #126, opened 2015-07-09
Enable the use of babel configuration for Gulp.

JacopKane:
Enable the use of babel configuration for Gulp.

### DIFF
--- a/lib/tools/gulp.js
+++ b/lib/tools/gulp.js
@@ -12,7 +12,9 @@ var originalNodePath = process.env.NODE_PATH;
 module.exports.niceName = 'gulp';
 
 module.exports.isEligable = function (cwd) {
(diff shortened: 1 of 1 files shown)


noseglid:
Thanks! :)

JacopKane:
I didn't test it that much but it's also working with included tasks (my default task is under tasks/default.js for example):

```javascript
import env from './build/env';
import requireDir from 'require-dir';
import gulp from 'gulp';
import LoadPlugins from 'gulp-load-plugins';

const plugins = LoadPlugins();

requireDir(`./${env.DIR_BUILD}/${env.DIR_TASK}`);

plugins.npmScriptSync(gulp);
```

Also not sure what to do with more advanced task parsing support right now though :neutral_face:

PR #130, opened 2015-07-11
Fix Issue #129: stop hangs

### LINKED ISSUE #129: [Bug] Can't abort build, if command starts subprocesses
The following is only tested on windows.

## Problem
If I start coffee.cmd as a custom tool to build my project, it starts nodejs as a subprocess to run the coffee code. When this code hangs and I want to abort it, atom-build hangs with
> Build Process termination imminent...

because it only kills the parent process, but not the child nodejs process.

## Fix
use tree-kill to kill the whole process tree

MackieLoeffel:
Fixes #129 using tree-kill

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -379,7 +379,7 @@ module.exports = {
     }.bind(this));
 
     try {
-      kill(this.child.pid);
+      kill(this.child.pid, 'SIGKILL');
     } catch (e) {
       /* Something may have happened to the child (e.g. terminated by itself). Ignore this. */
     }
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -129,7 +130,7 @@ module.exports = {
       this.customFileWatcher.close();
     }
     if (this.child) {
-      this.child.kill('SIGKILL');
+      kill(this.child.pid);
```
noseglid: I think this should still send 'SIGKILL' to the process, to forcefully terminate (this happens the second time the user aborts the build).

It appears to be easy with the `tree-kill` npm, just add 'SIGKILL' as second argument.

noseglid:
Had to merge this manually due to a conflict! Thanks for your contribution!

PR #131, opened 2015-07-11
rust + cargo: Match failing testcases

colin-kiegel:
It should only match if a test 'panicked' in local code (e.g. via macro panic!, assert!, assert_eq!, unreachable!, unimplemented!, etc.) but not if the error occurred in the std-library (like Err(()).unwrap());

The distinction is done by only matching *relative* paths (not sure if this is 100% accurate - but working for me).

```rust
#[test]
pub fn A() {
    unimplemented!(); // matched
    // thread '...' panicked at 'not yet implemented', relative/path/to/my/file.rs:123
}

#[test]
pub fn B() {
    assert!(false); // matched
    // thread '...' panicked at 'assertion failed: false', relative/path/to/my/file.rs:123
}

#[test]
pub fn C() {
    let x : Result<(),()> = Err(());
    x.unwrap(); // *not* matched
    // thread '...' panicked at 'called `Result::unwrap()` on an `Err` value: ()', /home/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libcore/result.rs:729
}
```

### DIFF
--- a/lib/tools/Cargo.js
+++ b/lib/tools/Cargo.js
@@ -14,20 +14,20 @@ module.exports.settings = function (path) {
     exec: 'cargo',
     args: [ 'build' ],
     sh: false,
-    errorMatch: '(?<file>[^\\.]+.rs):(?<line>\\d+):(?<col>\\d+):'
+    errorMatch: '(?<file>.+.rs):(?<line>\\d+):(?<col>\\d+):'
   },
   {
     name: 'Cargo: test',
     exec: 'cargo',
     args: [ 'test' ],
     sh: false,
-    errorMatch: '(?<file>[^\\.]+.rs):(?<line>\\d+):(?<col>\\d+):|thread \'[^\\\']+\' panicked at \'[^\\\']+\', (?<file>[^\\/][^\\:]+):(?<line>\\d+)'
+    errorMatch: '(?<file>.+.rs):(?<line>\\d+):(?<col>\\d+):|thread \'[^\\\']+\' panicked at \'[^\\\']+\', (?<file>[^\\/][^\\:]+):(?<line>\\d+)'
   },
   {
     name: 'Cargo: run',
     exec: 'cargo',
     args: [ 'run' ],
     sh: false,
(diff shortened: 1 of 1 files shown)


colin-kiegel:
Important: This introduces an issue that regular matches can not be evaluated correctly ...

Looks like a pattern like this only works correctly if the second alternative matches
      '(?<file>A)|(?<file>B)' 

If the first alternative matches I get 'Did not match any file. Don't know what to open.'

Looks like it does match, but cant evaluate the match ...

noseglid:
It appears as the XRegExp library does not support named groups together with the `|` operator.
I think some syntax similar to what you're describing is the best way to address this shortcoming.

I'm a tad reluctant at merging this at the moment as it would break error matching for those who had it working before. I will keep this PR open as a reminder to get that new syntax done.

Thanks for the effort!

MackieLoeffel:
I encountered the same problem and implemented the proposed syntax of @colin-kiegel in a branch.
My idea was to support both variants so these both work and configs of old users wouldn't be broken:
```javascript
errorMatch: [
  '(?<file>patternA)',
  '(?<file>patternB)'
]

errorMatch: '(?<file>pattern)'
```
@noseglid: How do you think about this?

I would create a pull request, but the branch, on which I implemented this, is based of #137, so merging it would introduce all the changed made by it.

noseglid:
@MackieLoeffel That is excellent. It wouldn't break backward compatibility, and allows for multiple regexes. Let's get #137 finalized and then we can look at that branch! Thanks for your effort!

colin-kiegel:
Great. I updated this PR to reflect the planned multipattern syntax for error matching. :-)

So it should be possible to merge this PR as soon as #137 and the new branch of MackieLoeffel are integrated.

noseglid:
This completely slipped my mind. I'm sorry :disappointed: . 

I've merged this to the stand-alone rust/cargo package noseglid/atom-build-cargo#1. Thanks for this!

PR #138, opened 2015-07-15
Add Marked.app and JPM support.

bwinton:
Ideally, this would be a couple of plugins for atom-build, but I'm not sure how to do that without littering files all over my filesystem…  (And in particular, I don't think it's possible to have a per-file build target with the current `.atom-build.json` spec.)  All the built-in tools seem to have a common API, so it should be possible…  If you plan to steal the code from [AtomLinter](https://github.com/AtomLinter/Linter/wiki/Linter-API), then please let me know, and I’ll be happy to re-write these as plugins to atom-build instead.  :wink:


noseglid:
I would love this support, but I think we should focus our efforts right now on the service API you PR'd, and this can be plugins rather. What do you think?

bwinton:
Yeah, totally!  This was just a hack back when I thought the service API would be difficult.  :smile:
(I'm going to close it, to save you the trouble.)

noseglid:
Thanks! :)

PR #139, opened 2015-07-15
Use a service instead of hard-coding all the tools.

bwinton:
It ended up being pretty easy to use a Service-based architecture, and I had some free time, so I did this.  :smile:

See https://github.com/bwinton/atom-build-jpm or https://github.com/bwinton/atom-build-markdown for example service providers.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -13,6 +13,7 @@ var BuildView = require('./build-view');
 var GoogleAnalytics = require('./google-analytics');
 var ErrorMatcher = require('./error-matcher');
 var tools = require('./tools');
+var extra_tools = [];
 
 function BuildError(name, message) {
   this.name = name;
@@ -185,7 +186,7 @@ module.exports = {
           throw new BuildError('No active project', 'No project is active, don\'t know what to build...');
         }
 
-        return _.filter(tools, function (tool) {
+        return _.filter(tools.concat(extra_tools), function (tool) {
           return tool.isEligable(path);
         });
       })
@@ -437,5 +438,14 @@ module.exports = {
     } else {
       this.buildView.reset();
     }
+  },
+
+  consumeBuilder: function (builders) {
+    if (!(builders instanceof Array)) {
+      builders = [builders];
+    }
+    builders.forEach(function (builder) {
+      extra_tools.push(builder);
+    });
   }
 };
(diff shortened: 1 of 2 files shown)


noseglid:
Not bad! I will have a look at it as soon as I have some time over!

noseglid:
I'm curious. Why did you choose to have `atom-build` consume a service which is provided by the different packages? Would it be more natural the other way around?

bwinton:
It's a good question.  Mostly I was basing it off the [AtomLinter](https://github.com/AtomLinter/Linter/blob/master/package.json#L15) code, which does things this way.  My other thought was that since atom-build is the thing that's in control, calling the other packages, it's kind of using their linting services, but that's really more of an after-the-fact justification than a reason…  The other main reason is that it seemed to work this way, so I didn't bother trying to get it working the other way.  :wink:

bwinton:
(One thing I just noticed is that `consumeBuilder` will need to return a `Disposable` that removes the builders from `extra_tools`, so that we don't try to use them when someone disables or uninstalls the extra packages…  I'm not sure how that would work if atom-build was providing the service.)

noseglid:
After looking at how the Linter does it, it seems best to have it this way.

I think we should fix the `Disposable` thing as well. It seems fairly easy, but we need to assign an `id` in the extra_tools array (making it an object) so we can remove it.

Also, how should a build tool (e.g. your markdown plugin) react if `atom-build` is removed? Would it even get notified?

noseglid:
I ported the `gulp` tool to a [repository of its own](https://github.com/noseglid/atom-build-gulp) as well. Works really well and is an excellent modularization of atom-build!

bwinton:
(Just as a note, I'll probably get to this tomorrow afternoon, or sometime Saturday, if you don't fix it first.)

bwinton:
> I think we should fix the Disposable thing as well. It seems fairly easy, but we need to assign an id in the extra_tools array (making it an object) so we can remove it.

Fixed.  Since I believe the objects will be unique, I'm just using lodash to remove them (and I switched to using lodash to add them, for consistency).

> Also, how should a build tool (e.g. your markdown plugin) react if atom-build is removed? Would it even get notified?

It doesn't get any notifications, but it never gets called, so it all works out.  (It will get re-registered when `atom-build` is re-installed, and magically start working. :smile: )

noseglid:
This is great! Thank you for this!

bwinton:
You're very welcome.  And thank you for making atom-build` in the first place, so I didn't have to!

PR #142, opened 2015-07-21
[Feature] Multiple regexes

MackieLoeffel:
Allows to supply a list of regexes in errorMatch. See discussion in #131.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -349,7 +349,11 @@ module.exports = {
             }.bind(this), 1000);
           } else {
             var t = this.targets[this.activeTarget];
-              this.errorMatcher.set(t.errorMatch, this.replace(t.cwd), this.buildView.output.text());
+            var errorMatch = t.errorMatch;
+            if(!(errorMatch instanceof Array)) {
+              errorMatch = [errorMatch];
+            }
+            this.errorMatcher.set(errorMatch, this.replace(t.cwd), this.buildView.output.text());
             if (atom.config.get('build.scrollOnError')) {
               this.errorMatcher.matchFirst();
             }
--- a/lib/build.js
+++ b/lib/build.js
@@ -362,7 +362,7 @@ module.exports = {
           } else {
             var t = this.targets[this.activeTarget];
             var errorMatch = t.errorMatch;
-            if(!(errorMatch instanceof Array)) {
+            if (!(errorMatch instanceof Array)) {
               errorMatch = [errorMatch];
             }
             this.errorMatcher.set(errorMatch, this.replace(t.cwd), this.buildView.output.text());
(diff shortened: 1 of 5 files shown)


noseglid:
Wow, you're fast :+1:  I will take a look at this as soon as possible!

lib/build.js
```diff
@@ -361,7 +361,11 @@ module.exports = {
             }.bind(this), 1000);
           } else {
             var t = this.targets[this.activeTarget];
-            this.errorMatcher.set(t.errorMatch, this.replace(t.cwd), this.buildView.output.text());
+            var errorMatch = t.errorMatch;
+            if (!(errorMatch instanceof Array)) {
```
noseglid: I think this could be handled in the `set` function of `ErrorMatcher`. Also, it's probably cleaner to write `var errorMatch = (errorMatch instanceof Array) ? errorMatch : [ errorMatch ]`.

lib/error-matcher.js
```diff
@@ -77,12 +90,22 @@ module.exports = (function () {
...
+    for (var i in regex) {
+      var curRegex = null;
+      try {
+        curRegex = XRegExp(regex[i]);
+      } catch (err) {
+        this.emit('error', 'Error parsing regex:\n' + err);
+      }
+      if (curRegex !== null) {
+        this.regex.push(curRegex);
+      }
```
noseglid: Hmm, I think this can be better done with `map`. Something like this (untested):

```javascript
this.regex = _.map(regex, function (r) {
  try {
    this.regex.push(XRegExp(r));
  } catch (err) {
    this.emit('error', 'Error parsing regex: ' + err);
  }
});
```

lib/error-matcher.js
```diff
@@ -77,12 +90,22 @@ module.exports = (function () {
...
     }
+    if (this.regex.length === 0) {
+      this.regex = 0;
+    }
+
```
noseglid: Not sure why you would want to do this? Seems like just leaving it an empty array will also cause the loop on row `64` not to execute. Besides, `this.regex` will always be an array as well.

noseglid:
Sorry for taking so long to get around to this. I've been away for some time during the summer.

I've added some comments.

Also, it would be really neat with some specs for this, and note in the `README.md` file..

lib/error-matcher.js
```diff
@@ -78,22 +91,23 @@ module.exports = (function () {
...
-      this.regex = null;
-      return this.emit('error', 'Error parsing regex:\n' + err);
-    }
+    regex = (regex instanceof Array) ? regex : [ regex ];
+
+    this.regex = _.compact(regex.map(function(r) {
+      try {
+        return XRegExp(r);
+      } catch (err) {
+        this.emit('error', 'Error parsing regex:\n' + err);
```
noseglid: this is not bound here. You need to add `.bind(this)` after the last `}` in the function declaration.

noseglid:
I fixed the test error and merged this. Thanks you for this contribution!

PR #143, opened 2015-07-22
Comparison was somehow case sensitive?

Flonker:
Got BuildError "No active project" because path wasn't matched correctly.
Narrowed it down to being a case sensitive comparison.

Left side of the comparison was "C:\temp\file.js" while right was "c:\temp\file.js" so just different case on the drive letter.

This small change fixed the problem.


noseglid:
This solution is no good. On many filesystems (ext3, ext4, xfs, Optional HFS) the filenames are indeed case sensitive. On windows it is not.

I am not sure how this would be best solved, but the proposed solution is not viable.

PR #144, opened 2015-07-25
[Feature] Fail build on error match

MackieLoeffel:
Currently, if the build-tool exits with exitcode 0, the build is said to have succeeded. This is a problem, if the build-tool outputs an error, which would be recognized by errorMatcher, but also exits with 0. This PR changed the condition for a successfull build to also include, that there were no errors found by errorMatcher.

@noseglid What do think of this? Should this be the default or maybe add a flag to .atom-build.json, which enables this behavior?

This PR includes #145 (Travis build fails otherwise).


MackieLoeffel:
@noseglid Can you tell me, why the last [travis build](https://travis-ci.org/noseglid/atom-build/builds/75718707) is failing? It says, "atom.workspace" is null at some point and I have no idea, how this could happen. Since the tests are not running under windows, I can't test this on my machine.

noseglid:
I'm not sure... I thought it could've been some intermittent issue, but it doesn't run for me either.

I'll look into it on my local environment. The error doesn't really seem to have anything to do with the changes you've made. I think it's more likely that an existing race condition or similar is seen here.

Regarding windows, I am painfully aware. I have an idea on how to fix it. The problem is mainly that the specs use `.atom-build.json` files which contain commands such as `dd` or `printf` which runs on *nix, but not windows. My plan is to move all build tools (except custom `.atom-build.json`) out of this project and instead use the service API to provide different tools. At the same time change all the specs so they're platform compatible.

MackieLoeffel:
Maybe the test about Ansi-colors fails and travis complains in a very strange way. By the way, I didn't even know until yesterday, when I looked into the tests, that ansi-colors are supported. The errorMatcher removes them, when _parse is called, which is called after every build on this PR (before it was after every failed build). I have an idea, how to fix this, but it would restructure some code. When I find the time, I'll create a PR.

The missing-commands-problem is solvable by having the needed executables in PATH, when running the specs, e.g. by starting Atom from cygwin (and changing /bin/sh in build.js to the correct one, e.g. C:\cygwin64\bin\sh). The tests don't work for me for the following reasons:
 - A normal user cannot create symlinks in windows (only a admin can)
 - I don't have have the needed tools (grunt/...) installed
 - scrollTop is one pixel of in [this test](https://github.com/noseglid/atom-build/blob/1c20f28fb9b1fab7276360da6c1fbfbe35838373/spec/build-error-match-spec.js#L344), 101 in the test vs 100 on my machine.

MackieLoeffel:
@noseglid Can you tell me, why travis fails here? If I make some windows-related adjustments all tests pass on my machine.

noseglid:
Without testing it closer, could it be that you changed `this.buildView.output.text()` to  `this.buildView.text()` and there are some follow-up errors ?

noseglid:
Also, you should throw in a test to verify that it actually fails the build even if command executes successfully if error matcher matches.

MackieLoeffel:
Good spot! I've fixed it and added a testcase.

noseglid:
I did some digging in this. I think this is a race condition which existed previously, but became visible due to your changes since it probably changes timing a little.

The problem is that the package is deactivated, and if a child process exists, it will kill it. This will cause the child process to emit a `close` event, where we call the `replace` function which in turn calls `getActiveTextEditor`.

Atom has shut down at this point so there is no active text editor, thus it returns null.

Disabling all events before issuing kill in deactivate seems to fix it. Please try to add this on row 152 (inside the `if` case in deactivate: `this.child.removeAllListeners();`.

MackieLoeffel:
Added. Would be great, if this fixes all these strange travis errors.

noseglid:
Hmm, that didn't seem to fix it fully. At least the error message disappeared, which is good.
There has to still be some weird condition added by your code. It's not apparent to me what it is...

MackieLoeffel:
Do you know, when the package is deactived and actived during the tests? Before and after each test? Before and after each describe or each file? Can a test influence a following test or is the whole environment reseted? I can't find the relevant documentation.

noseglid:
I'm not certain. `beforeEach` call `atom.packages.activate('build')` which returns a promise that we wait for. But I think it is activated prior to this, we just wait it out.

Deactivation seems to happen "behind the scenes" as well.

noseglid:
I found another thing. Try setting `this.child = null`, in that same if-statement in `deactivate`. It appears as even though the package is deactivated and activated, `this` remains. And I noticed that when it fails, `this.child` is set.

I tried it on your branch locally and every test works if `this.child` is set to `null` in the deactivate.

MackieLoeffel:
This seems to fix it (or the race condition didn't occur this time). Still have no idea, what the problem was, and how setting `this.child` to `null` fixes it, but great find!
The tests work on my machine with and without this line.

noseglid:
I investigated further. It appears as the next build is never triggered as a consequence of this.
If a build is running, it will wait for it to abort (a `close` event from child). This never seems to come (even without the `removeAllListeners` change we made). I'm not sure why.

Either way, this seems to do the trick! Thanks for taking the time to debug this with me!

PR #145, opened 2015-07-25
[Bugfix] Parse on errorMatcher.set / handle no errorMatcher correctly

MackieLoeffel:
This PR fixes two bugs:
 - Errors are now highlighted, when the build is finished, not when the user goes to the first error
 - no error matching is performed, when no errorMatcher is set

### DIFF
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -24,6 +24,10 @@ module.exports = (function () {
   util.inherits(ErrorMatcher, EventEmitter);
 
   ErrorMatcher.prototype._gotoNext = function () {
+    if (0 === this.currentMatch.length) {
+      return;
+    }
+
     this._goto(this.currentMatch[0].id);
     this.currentMatch.push(this.currentMatch.shift());
   };
@@ -86,6 +90,8 @@ module.exports = (function () {
     this.cwd = cwd;
     this.output = output;
     this.currentMatch = [];
+
+    this._parse();
   };
 
   ErrorMatcher.prototype.match = function () {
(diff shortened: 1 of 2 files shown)


noseglid:
I tried to rebase this and run the test but I can't get it to work. Do you have time to look into this?

MackieLoeffel:
Done.
I also noticed, that the specs fail in many ways on windows, even if I start atom from cygwin. If I have time, I may look into fixing this.

lib/error-matcher.js
```diff
@@ -25,12 +26,25 @@ module.exports = (function () {
...
+
     this._goto(this.currentMatch[0].id);
-    this.currentMatch.push(this.currentMatch.shift());
   };
 
   ErrorMatcher.prototype._goto = function (id) {
     var match = _.findWhere(this.currentMatch, { id: id });
+    if (!match) {
+      throw new Error('Can\'t find match with id ' + id);
+    }
```
noseglid: Doesn't seem like this can ever occur. In the `_gotoNext` function it is protected by a check for `0 === this.currentMatch.length`, and in the `matchFirst` function it is checked that there actually is a first match.
(If I missed some case where it actually happens, it should `emit` the error rather than throwing).

MackieLoeffel: You're correct, that it schouldn't happen. It's more like defensive programming, if the other functions change in the future. But you are also correct, that it should emit an error.

noseglid:
Thank you so much for this! Once again, sorry for taking so long to review this

PR #147, opened 2015-07-27
Updated error matching in README.md

gogoprog:
Hello,

I updated the regex to match the full path.
Also I had to replace the ^ by \n to make it functional.


noseglid:
Cool. Thanks!

PR #150, opened 2015-08-07
Option to customize output panel font size added.

Faresis:
Default font size in the build output panel is not customizable. This is very annoying when working from small screens e.g. 14 inch. Font is very small and very hard to read. While font in the main editor area can be customized. This should be fixed.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -23,6 +23,7 @@ module.exports = (function() {
     atom.config.observe('build.panelVisibility', this.visibleFromConfig.bind(this));
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
     atom.config.observe('build.minimizedHeight', this.heightFromConfig.bind(this));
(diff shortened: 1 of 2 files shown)


noseglid:
Is there a reason you can't do this in your custom CSS file?

Faresis:
Hi,

I am not sure how it should be handled.
I can create a .css file. But how it will be bound to the build-view?

Also, just woundering. Is that my chagnes broked CI build? Or something
else?

Please advise.

Thanks!

2015-08-07 9:10 GMT+03:00 Alexander Olsson <notifications@github.com>:

noseglid:
Open the command palette (`cmd-alt-p` or `ctrl-alt-p`). Type `stylesheet`. Add something like this:

```css
.build .output {
  font-size: 20px;
}
```

Modify as you want.

Faresis:
Yes, you are right. It is much more efficient way to handle the issue.

Thank You for the advice.

2015-08-07 10:50 GMT+03:00 Alexander Olsson <notifications@github.com>:

PR #155, opened 2015-08-15
[Bugfix] Not all windows programs use \r\n as line ending

MackieLoeffel:
Some programs in windows only use \n as newline, e. g. coffee. Without this fix, the output of such a program is only displayed, when it has finished running. Since both \r\n and \n end with \n, this fix should not cause problems.


noseglid:
Thanks!

PR #158, opened 2015-08-18
[Bugfix] Don't insert additional newlines

MackieLoeffel:
Code like the following could cause additional linebreaks in the build-window:
```coffeescript
process.stdout.write("same line\nsame");
setTimeout (-> process.stdout.write("line\n")), 1000
```
This PR fixes this bug, by removing the buffer and simply appending to the first `li` in the output instead of adding new `li`.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -186,7 +186,7 @@ module.exports = (function() {
     this.titleText.text('Aborted!');
   };
 
-  BuildView.prototype.append = function(data, force) {
+  BuildView.prototype.append = function(data) {
     this.outputContent.append(this.a2h.toHtml(_.escape(data)));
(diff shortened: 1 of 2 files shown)


noseglid:
I actually noticed this today! Well spotted! My initial thought was to fix the buffering (e.g. by substringing everything up to `\n` and output that. Your solution seems cleaner though. Makes me wonder, do we really need the `ol` and the `li` ?

MackieLoeffel:
I didn't exactly know, if there was another reason for the `ol` and `li`, so I left them in (and I was to lazy to remove them ;) ). But I don't really see a reason, why they are needed anymore.

noseglid:
I think they can be removed. But let's do that in another PR! Thanks for looking into this!

PR #159, opened 2015-08-18
[Bugfix] ansi colors

MackieLoeffel:
Until now the colors in the output were removed, when the result in the buildpanel was parsed. This PR fixes this problem.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -226,7 +226,7 @@ module.exports = (function() {
     var tagIndex = 0; // index in tags
     var openTags = []; // currently open tags
 
-    // the chrarcters ', " and ` are escaped by lodash
+    // the characters ', " and ` are escaped by lodash
     // but, if we read the html from the build-panel, these characters aren't escaped anymore
     // so we have to write our own escape function, which only escapes the characters,
     // which stay escaped in the html (&, <, >)
@@ -259,6 +259,19 @@ module.exports = (function() {
 
       return htmlOutput.substr(prevHtmlIndex, htmlIndex - prevHtmlIndex);
     };
+
+    var openAllTags = function () {
+      for (var i in openTags) {
+        newOutput += openTags[i][0];
+      }
+    };
+
+    var closeAllTags = function () {
+      for (var i = openTags.length - 1; i >= 0; i--) {
+        newOutput += '</' + openTags[i].tagname + '>';
+      }
+    };
+
     for (var linkIndex in links) {
       var link = links[linkIndex];
 
@@ -271,23 +284,15 @@ module.exports = (function() {
 
       // close all openTags and reopen them inside the <a>-tag
       // otherwise the style of the <a> would override the style of the enclosing spans
-      for (var i in openTags) {
-        newOutput += '</' + openTags[i].tagname + '>';
-      }
+      closeAllTags();
       newOutput += util.format('<a class="%s" id="%s">', link.type, link.id);
-      for (i in openTags) {
-        newOutput += openTags[i][0];
-      }
+      openAllTags();
 
       newOutput += advance(link.index + link.length);
 
-      for (i in openTags) {
-        newOutput += '</' + openTags[i].tagname + '>';
-      }
+      closeAllTags();
       newOutput += '</a>';
-      for (i in openTags) {
-        newOutput += openTags[i][0];
-      }
+      openAllTags();
     }
     newOutput += htmlOutput.substr(htmlIndex);
 
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -2,6 +2,8 @@
 'use strict';
 
 var _ = require('lodash');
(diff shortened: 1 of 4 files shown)


noseglid:
Hmm, all this manual handling of the textbuffers make me cringe. Let me propose an alternate solution.

If we keep the clear-text buffers (without any HTML) somewhere accessible to both `errorMatcher` and `buildView`, then they could do whatever they want.

Now, `errorMatcher` does some replacing in `build-view` which means it needs some basic knowledge of it. This is not a good pattern IMO. Core `build` should do all interaction.

The `replace` function in buildView should also not require the full output. I think a better interface would be a string, and what it should be replaced with. buildView needs to keep the clear text buffer so it can do the replace in that, then re-render the view (rather than errorMatcher doing everything as it is now. That's what making it so difficult I think).

Thoughts?

MackieLoeffel:
I'm also not quite happy with the solution, as it stands.

But I don't fully understand, what you are proposing. Should there be a copy of the clear text output in `build`? This wouldn't be very useful for `errorMatcher`, because it must insert `<a>` tags in the output, but it wouldn't make sense to insert html-tags in the clear text output.

An idea would be to change the `replace` function to an `insertLink` function, which is then called by `errorMatcher` directly or indirecty through `build` for each found match. So the second part of `_parse` would move to `build-view`, which would make the code cleaner.

But I can't see a solution, which gets rid of the code in `_parse`, because the functionality is needed (we need to close & reopen all spans around html-tags for the color to be correct).

noseglid:
I guess what I'm trying to say in a very convoluted way is that I'm unhappy with the interface between buildView and errorMatcher. I think the fact that errorMatcher needs to take into account that buildView "might" have changed some content to be wrapped in HTML tags mean that buildView leaks too much of it's internals.

I think errorMatch can still emit the `replace` signal. But it should just say which string, and what it wants to do with it. It should *not* include HTML, but rather tell buildview that it wants a tag of this string in some way. buildView then knows that the output actually is HTML and does the appropriate thing.

For buildView to be able to achieve this, I think it needs to keep/have access to the clear text so it can add the `<a>` tag even if that part is colored in some way.

MackieLoeffel:
What do think about the changes in my last three commits?
`errorMatcher` doesn't need to know anything about the internals of `buildView`, exactly as you propose. It simply emits an event, when it has parsed the output. `build` then takes the matches and asks `buildView` to insert a link for each of them. So `buildView` and `errorMatcher` are independent, there is only a little glue code in `build`, which connects the two.

noseglid:
So, all the code in buildView is why I want it to have the clear text output. Now when inserting links it needs to take into account that there may exist spans etc. If we had the cleartext, it would be easy to first insert the links, then run the whole thing through `a2h` and we'd be set. no?

MackieLoeffel:
I don't see an easy solution, but I could miss something. The problem I see with inserting links and then running `a2h` is twofold:
* `errorMatcher` doesn't want to run on the raw output, but on the output with _ansi escape codes removed_, otherwise these escape codes would interfere with error matching. So you would have to find the positions in the raw output, that correspond to the matches. 
* even if we could insert the links in the correct places in the raw output, `a2h` would not necessarily generate vaild html, for example:
```javascript
> a2h.toHtml('\x1B[31m<a>dasd\x1B[0mdd</a>')
'<span style="color:#A00"><a>dasd</span>dd</a>'
```
The logic in `buildView` is fixing such invalid html.

Do you see a nice solution to these problems?

noseglid:
I'll have to do some experiments.

I'm really reluctant to merge all those string magics. I foresee a world of weird bugs which might come in the build output. I'm just not prepared to maintain it.

MackieLoeffel:
I hope, that the code isn't too complicated to understand, but if you have questions regarding it, I'll gladly answer them. And I'm quite confident, that the code is working as expected, but you can try some edge cases on your own.

Of course you have ot maintain the code later on, so it is your decision, how you fix this bug.

noseglid:
I've been working on this a bit and I have yet to find a solution to the "adding links" and "replacing ansi colors" simultaneously. I might very well lean towards not supporting ansi colors when error matching is performed (as it is now).

Regarding the interface discussion I mentioned, I've pushed it to [a branch named `build-view-interface`](https://github.com/noseglid/atom-build/compare/build-view-interface).

MackieLoeffel:
I like, what you did on the `build-view-interface` branch (except for the two bug, which I pointed out there, but all in all I think it's a good direction).

One thing I learned at my current workplace is, that the user is more important than the developer. So if there is a nice feature, it should be implemented, even if there is no "nice" implementation.
It should be possible to integrate the code in the `insertLinks` function of this PR into the `_render` function of the `build-view-interface` PR, so all tricky code is in this single function. But these are just my two cents.

jhasse:
+1 for ansi colors. I have "Automatically scroll on build error" activated, so colors always get removed for me :(

PR #167, opened 2015-08-23
Build view interface

noseglid:
The interfaces between `build`, `error-matcher` and `build-view`
is improved to decouple the modules. `error-matcher` no longer
have any knowledge of how the matching it performs is actually
used, and `build-view` no longer knows what the links it creates
are used for.

All together: better design.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -204,6 +204,11 @@ module.exports = (function() {
   };
 
   BuildView.prototype.link = function(text, id, onClick) {
+    if (_.findWhere(this.links, { text: text })) {
+      /* This text already has a link to it. Ignore it. */
+      return;
+    }
+
     this.links.push({
       id: id,
       text: text,
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -3,7 +3,6 @@
 
 var _ = require('lodash');
 var View = require('atom-space-pen-views').View;
-var $ = require('atom-space-pen-views').$;
 var cscompatability = require('./cscompatability');
 var GoogleAnalytics = require('./google-analytics');
 var Convert = require('ansi-to-html');
@@ -20,6 +19,8 @@ module.exports = (function() {
     this.a2h = new Convert();
     this.monocle = false;
     this.starttime = new Date();
+    this.buffer = new Buffer(0);
+    this.links = [];
 
     atom.config.observe('build.panelVisibility', this.visibleFromConfig.bind(this));
     atom.config.observe('build.monocleHeight', this.heightFromConfig.bind(this));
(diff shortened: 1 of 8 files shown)


lib/build-view.js
```diff
@@ -182,22 +185,35 @@ module.exports = (function() {
     this.titleText.text('Aborted!');
   };
 
+  BuildView.prototype._render = function() {
+    var string = this.a2h.toHtml(_.escape(this.buffer.toString('utf8')));
+    _.forEach(this.links, function (link, index) {
+      string = string.replace(link.text, '<a id="' + link.id + '">' + _.escape(link.text) + '</a>');
+    });
```
MackieLoeffel: When I tested this, I found two bugs:
 * you must replace `_.escape(link.text)` instead of `link.text`, because `string` is escaped and `link.text` is not
 * the replace fails, when there are two `link.text`, which are the same (`<a>` is wrapped around the first match twice and not around the second)

noseglid: Thank you for the feedback! You are absolutely correct. I have fixed this in the next commit!

Any other issues you can foresee with this? Is there anything I have forgotten?

noseglid:
Thanks for the input! I've fixed the first two issues.

Regarding the case where ANSI codes are part of the matched text: If the output from the build command contains some non-printable characters `atom-build` does not know whether they are of importance or not. If they are expected there, they also should be part of the regular expression. For example (nodejs):

```javascript
var s = "chinese \u6f22\u5b57 colored \x1B[31mstring\x1B[m clear";
console.log(s);

console.log(s.match(/chinese .. ([0-9a-z \x1b\x5b\[]*)/));
```

Granted, if the filepath is cut of in the middle with an escape code, it may be difficult. I think the user would have to resolve to disable colored output in that case.

MackieLoeffel:
I really don't like the idea, that the user should deal with ANSI codes himself. I can't imagine a case, where the user doesn't want to match a text if it contains escape codes, but wants to match it, if it doesn't.
ANSI escape codes are not intuitive to understand and I'm sure, that many users will be glad, if they don't have to deal with them.
It would be quite frustrating for a user, if he tries to get a custom matcher working, but doesn't manage to achieve it, because he has never heard of ANSI escape codes.
It should be possible to port the code from #159 to this new architecture, I can look into it, if you want to.

noseglid:
I totally agree with you. It is not optimal.

Another way to go about it could be with an option of matching on output "as-is" or remove certain known structures (such as ANSI color escape codes). If opting in for this, the output window would be without colors.

There are several reasons I'm reluctant to merge #159. Complexity of the implementation is one. I am certain there will be more cases which doesn't really work with the implementation at hand.
The other is that there could be more things we might want to do with the output. Could be replacing "http" with clickable links. All this becomes exponentially more difficult with manual string replacement.

noseglid:
I'll merge this, and if there's a lot of people having issues with this, we will have to reconsider! If no one reports any problems, it's probably fine!

Issue #173, opened 2015-09-09
Uncaught TypeError: type RegExp expected


bodocabdd:
So, i have this project using a custom .atom-build.json, using cake cmd to join coffee files, among other tasks. My problem is, it was working really fine before some atom and atom-build updates. And now, every time I try to build with an error in my code, the package throws this Exception below and I have to restart atom to be able to trigger build again. I would love to help you and fix this, but I lack the time right now. =(

**Steps**
1. A project thats builds coffee with cake.
2. This .atom-build.json:
```JSON
{
    "cmd": "cake",
    "args": ["sbuild"],
    "sh": true,
    "cwd": "{PROJECT_PATH}"
}
```

**Atom Version**: 1.0.11
**System**: Mac OS X 10.10.5
**Thrown From**: [build](https://github.com/noseglid/atom-build) package, v0.46.0

### Stack Trace

Uncaught TypeError: type RegExp expected

```
At /Users/myuser/.atom/packages/build/node_modules/xregexp/xregexp-all.js:141

TypeError: type RegExp expected
    at copy (/Users/myuser/.atom/packages/build/node_modules/xregexp/xregexp-all.js:141:19)
    at Function.self.exec (/Users/myuser/.atom/packages/build/node_modules/xregexp/xregexp-all.js:504:18)
    at Function.self.forEach (/Users/myuser/.atom/packages/build/node_modules/xregexp/xregexp-all.js:541:30)
    at ErrorMatcher.module.exports.ErrorMatcher._parse (/Users/myuser/.atom/packages/build/lib/error-matcher.js:80:15)
    at ErrorMatcher.module.exports.ErrorMatcher.set (/Users/myuser/.atom/packages/build/lib/error-matcher.js:127:10)
    at Object.<anonymous> (/Users/myuser/.atom/packages/build/lib/build.js:354:31)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:169:7)
    at maybeClose (child_process.js:997:16)
    at Socket.<anonymous> (child_process.js:1166:11)
```

### Commands

```
     -0:36.8.0 build:trigger (atom-text-editor.editor.is-focused)
  3x -0:28 editor:delete-line (atom-text-editor.editor.is-focused)
     -0:26.9.0 core:save (atom-text-editor.editor.is-focused)
     -0:26 core:backspace (atom-text-editor.editor.is-focused)
     -0:25.8.0 core:save (atom-text-editor.editor.is-focused)
[...]

PR #175, opened 2015-09-10
There, I fixed issue #173

bodocabdd:
This should do the trick. Not sure if this is the most elegant solution, but worked for me.
Hope it helps. ;)

Fixes issue #173

### DIFF
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -88,7 +88,7 @@ module.exports = (function () {
 
     var output = '';
     var lastEnd = 0;
-    for (var matchIndex = 0; this.currentMatch.length && matchIndex <= this.currentMatch.length; matchIndex++) {
+    for (var matchIndex = 0; this.currentMatch.length && matchIndex < this.currentMatch.length; matchIndex++) {
       var match = this.currentMatch[matchIndex];
 
       if (match.index < lastEnd) {
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -76,7 +76,7 @@ module.exports = (function () {
       match.id = 'error-' + regexIndex + '-' + i;
       this.push(match);
     };
-    for (var regexIndex in this.regex) {
+    for (var regexIndex = 0; this.regex.length && regexIndex < this.regex.length; regexIndex++) {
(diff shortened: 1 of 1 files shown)


lib/error-matcher.js
```diff
@@ -88,7 +88,7 @@ module.exports = (function () {
 
     var output = '';
     var lastEnd = 0;
-    for (var matchIndex in this.currentMatch) {
+    for (var matchIndex = 0; this.currentMatch.length && matchIndex <= this.currentMatch.length; matchIndex++) {
```
noseglid: This doesn't seem right. The index could never be equal to the length of the array.

noseglid:
I don't see how this code differs from what was there before?

lib/error-matcher.js
bodocabdd: Sorry. It was supposed to be < only, not <=. My bad. Didn't catch that.

bodocabdd:
It's different. You use ```for(var index in object)``` with array, and it was detecting "unique" proto function as enumerable property, which was triggering the error.

bodocabdd:
I guess you could use a ```for(var index of array)``` or ```array.forEach``` also.

noseglid:
I can see how this would be a problem if `this.regex` contained anything else than regular expressions. But what I don't get is how it can contain other stuff for you? It seems as there is something more happening here.

bodocabdd:
Hmm... I looked up my packages. Seems that the "atom-permute" deprecated package is causing the problem. "unique" is declared as a prototype function for Array. That explains why it is detected as enumerable property when you use ```for (key in array)```. Disabling the package solves the problem. But I still recommend you change those lines...

noseglid:
Ohh extending the prototype! It's forbidden! Forbidden I say! Uninstall that crap and throw it away :)

I've implemented a more defensive strategy in #177 . Would you mind testing if that works for you?

bodocabdd:
Already did, right after seeing that. ;) Never used atom-permute anyway.
Your strategy is very good, sir! ES6 FTW! I guess this closes the case.

Thank you very much for your time.

noseglid:
Thank you for taking the time to report and debugging this with me!

PR #176, opened 2015-09-10
Support a specified `errorFile` for no-file error messages

MadcapJake:
Allows service providers and users to provide an `errorFile` that is given to the ErrorMatcher if the match does not contain or does not match a file.  This allows ErrorMatcher to be used with any process that uses a single file or just error messages that do not specify a file.

This works for me with a few different projects of mine that use `atom-build` (including ones that do not specify `errorFile`).  However, running `apm test` is giving me a lot of timeout errors.  Any help in finding what's causing these timeouts would be appreciated as I'd love to get this feature merged!

I will add a test json file, update `build-error-match-spec.js`, and update docs once I fix these timeouts.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -350,7 +350,7 @@ export default {
     }).catch((err) => targetsView.setError(err.message));
   },
 
-  replace(value, targetEnv) {
+  replace(value = '', targetEnv) {
     const env = _.extend({}, process.env, targetEnv);
     value = value.replace(/\$(\w+)/g, function (match, name) {
       return name in env ? env[name] : match;
--- a/lib/build.js
+++ b/lib/build.js
@@ -29,7 +29,7 @@ export default {
     },
     buildOnSave: {
       title: 'Automatically build on save',
-      description: 'Autmatically build your project each time an editor is saved.',
+      description: 'Automatically build your project each time an editor is saved.',
       type: 'boolean',
       default: false,
       order: 2
--- a/lib/build.js
+++ b/lib/build.js
@@ -8,6 +8,7 @@ import kill from 'tree-kill';
 import Grim from 'grim';
 
 import SaveConfirmView from './save-confirm-view';
+import StatusBarView from './status-bar-view';
 import TargetsView from './targets-view';
 import BuildView from './build-view';
 import GoogleAnalytics from './google-analytics';
@@ -61,6 +62,13 @@ export default {
       default: true,
       order: 6
     },
+    selectTriggers: {
+      title: 'Selecting new target triggers the build',
+      description: 'When selecting a new target (through status-bar, cmd-alt-t, etc), the newly selected target will be triggered.',
+      type: 'boolean',
+      default: true,
+      order: 7
+    },
     monocleHeight: {
       title: 'Monocle Height',
       description: 'How much of the workspace to use for build panel when it is "maximized".',
@@ -68,7 +76,7 @@ export default {
       default: 0.75,
       minimum: 0.1,
       maximum: 0.9,
-      order: 7
+      order: 8
     },
     minimizedHeight: {
       title: 'Minimized Height',
@@ -77,7 +85,7 @@ export default {
       default: 0.15,
       minimum: 0.1,
       maximum: 0.9,
-      order: 8
+      order: 9
     },
     panelOrientation: {
       title: 'Panel Orientation',
@@ -85,7 +93,15 @@ export default {
       type: 'string',
       default: 'Bottom',
       enum: [ 'Bottom', 'Top', 'Left', 'Right' ],
-      order: 9
+      order: 10
+    },
+    statusBar: {
+      title: 'Status Bar',
+      description: 'Where to place the status bar. Set to `Disable` to disable status bar display.',
+      type: 'string',
+      default: 'Left',
+      enum: [ 'Left', 'Right', 'Disable' ],
+      order: 11
     }
   },
 
@@ -132,6 +148,8 @@ export default {
       });
     });
 
+    atom.workspace.onDidChangeActivePaneItem(() => this.updateStatusBar());
+
     this.errorMatcher.on('error', (message) => {
       atom.notifications.addError('Error matching failed!', { detail: message });
     });
@@ -170,6 +188,9 @@ export default {
       kill(this.child.pid, 'SIGKILL');
       this.child = null;
     }
+
+    this.statusBarView && this.statusBarView.destroy();
+
     clearTimeout(this.finishedTimer);
   },
 
@@ -189,6 +210,11 @@ export default {
       const realpath = fs.realpathSync(p);
       return textEditor.getPath().substr(0, realpath.length) === realpath;
     });
+  },
+
+  updateStatusBar() {
+    const activeTarget = this.activeTarget[this.activePath()];
+    this.statusBarView && this.statusBarView.setTarget(activeTarget);
   },
 
   cmdDefaults(cwd) {
@@ -286,6 +312,7 @@ export default {
 
         this.targets[p] = settings;
         this.targetsLoading[p] = false;
+        this.updateStatusBar();
       });
     });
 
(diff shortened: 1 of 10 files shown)


MadcapJake:
Well I guess I must not have my dev environment set-up properly, because the CI tests pass!

noseglid:
I'm not convinced this is the right way to go about it. It seems counter-intuitive that the settings should contain the file. Keep in mind that these settings are exactly the same if using an `.atom-build.json` file. It would create a very limited build file if you had to specify the exact file there...

MadcapJake:
You don't *have* to specify it, it's only for projects that work on individual files.  And you would use `{FILE_ACTIVE}` as it uses the `replace` method to get the active file and sets it to `errorFile`. Another way you could use this would be to have your process write to a log file and then `errorFile` could be the log file itself.

MadcapJake:
Plus, `.atom-build.json` can use the replacements too, so you don't have to specify the exact file everytime, all you need to do is have the active tab that you would like to use and then `build:trigger`.

MadcapJake:
If you have another route that you feel is more intuitive, I'd be happy to implement it, I just want this feature :smile:

lib/build.js
```diff
@@ -351,7 +351,11 @@ module.exports = {
             }.bind(this), 1000);
           } else {
             var t = this.targets[this.activeTarget];
-            this.errorMatcher.set(t.errorMatch, this.replace(t.cwd), this.buildView.output.text());
+            if (t.errorFile) {
+              this.errorMatcher.set(t.errorMatch, this.replace(t.cwd), this.buildView.output.text(), this.replace(t.errorFile));
+            } else {
+              this.errorMatcher.set(t.errorMatch, this.replace(t.cwd), this.buildView.output.text());
+            }
```
noseglid: I think `error-matcher` should take care of whether `errorFile` is set or not, rather than duplicating this line

spec/fixture/.atom-build.error-match-errorFile.json
```diff
@@ -0,0 +1,5 @@
+{
+	"cmd": "echo 'line:3,column:8' && return 1",
+	"errorMatch": "line:(?<line>\\d+),column:(?<col>\\d+)",
+  "errorFile": ".atom-build.json"
+}
```
noseglid: Indentation is weird in this file. Also, you could consider if you want to define this in the test instead of having a file:

```javascript
fs.writeFileSync(directory + '.atom-build.json', JSON.stringify({
  cmd: "echo 'line:3,column:8' && return 1",
  errorMatch: "line:(?<line>\\d+),column:(?<col>\\d+)",
  errorFile: ".atom-build.json"
}).;
```

I've been leaning more towards that lately as the whole test is encapsulated in that one `it`.

noseglid:
You might just have convinced me. :)

This option needs to be added to the `README.md` as well.

It definitely seems as something you did affected the specs. Not sure why...

MadcapJake:
Any ideas why error-matcher tests are failing?  I don't understand why they're timing out. :goberserk:

noseglid:
Would you mind merging master to this branch and try again?

noseglid:
There's likely some error introduced here. It appears as the test you wrote fails, and then there are follow up errors on that.

MadcapJake:
I'll take a look over it again.  I just don't understand why it's timing out rather than actually failing.

noseglid:
You are using `waitsFor` which spends up to 5000 ms waiting for some condition (e.g. that the build panel errors). If it doesn't error within 5000 ms, it will fail with a timeout.

noseglid:
Will you continue on this?

MadcapJake:
Actually yeah I will. I should be able to devote some time this evening or tomorrow morning.  Sorry for the long delay!

MadcapJake:
Will you rerun Travis?

MadcapJake:
I may have botched that rebase :worried:

MadcapJake:
I have no idea why there are timeouts, the test is setup exactly like the first errorMatch test.  I think my JS skills aren't strong enough to solve this one.  Sorry for wasting your time, you can close it if you want.

noseglid:
I found the issues for this. There were multiple. I will put the diffs here and describe the change, so you can include them in your PR:

```
diff --git a/lib/atom-build.js b/lib/atom-build.js
index e822d5ebe88cdc4eead7ebb1fea86aa0ffaa54b1..da8e6852b3e4c59a49922d3646bca108abbf6b7f 100644
--- a/lib/atom-build.js
+++ b/lib/atom-build.js
@@ -31,6 +31,7 @@ function createBuildConfig(build, name) {
     cwd: build.cwd,
     sh: build.sh,
     errorMatch: build.errorMatch,
+    errorFile: build.errorFile,
     keymap: build.keymap
   };
 }
```
When `atom-build.js` is creating build configuration, it just doesn't take the JSON file straight off. It only takes the keys it knows about, thus we must tell it to also take the `errorFile` key.

```
diff --git a/lib/error-matcher.js b/lib/error-matcher.js
index 99cbfa99d9b3a3cefc470502de57a159724b46d4..6fa9e7f09930b410dbafe51e35a77687a84c742f 100644
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -42,12 +42,9 @@ export default class ErrorMatcher extends EventEmitter {
     }
     this.currentMatch.push(this.currentMatch.shift());

-    let file = match.file;
+    let file = match.file ? match.file : this.errorFile;
     if (!file) {
-      if (!this.errorFile) {
-        return this.emit('error', 'Did not match any file. Don\'t know what to open.');
-      }
-      match.file = this.errorFile;
+      return this.emit('error', 'Did not match any file. Don\'t know what to open.');
     }
```
After this check, the matched file is expected to be in the variable `file`, not `match.file`. I also reduced lines of code by doing a ternary before checking if both is null.

```
diff --git a/spec/build-error-match-spec.js b/spec/build-error-match-spec.js
index 573b9cdf98ea38fdfa67bd8f13210dd97b9943cc..a3ec7348da1f45039f72a9770674c57d053bec91 100644
--- a/spec/build-error-match-spec.js
+++ b/spec/build-error-match-spec.js
@@ -137,10 +137,10 @@ describe('Error Match', () => {
       expect(workspaceElement.querySelector('.build')).not.toExist();

       fs.writeFileSync(directory + '.atom-build.json', JSON.stringify({
-        cmd: 'echo "line:3,column:8" && return 1',
+        cmd: 'echo "line:3,column:8" && exit 1',
         errorMatch: 'line:(?<line>\\d+),column:(?<col>\\d+)',
         errorFile: '.atom-build.json'
-      }));
+      }, null, '  '));

       waitsForPromise(() => specHelpers.refreshAwaitTargets());

@@ -163,8 +163,8 @@ describe('Error Match', () => {
         const editor = atom.workspace.getActiveTextEditor();
         const bufferPosition = editor.getCursorBufferPosition();
         expect(editor.getTitle()).toEqual('.atom-build.json');
-        expect(bufferPosition.row).toEqual(3);
-        expect(bufferPosition.column).toEqual(8);
+        expect(bufferPosition.row).toEqual(2);
+        expect(bufferPosition.column).toEqual(7);
       });
     });
```
First, when serializing with `JSON.stringify`, the json will be minified, which means there is only one row which is very long. Using the [replacer and space](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) options, we can get the JSON we expect.

Second, Atom counts lines and cols starting at 0 (while error matcher starts at 1), so the matched row-col must be reduced with 1.

Hopefully this makes sense and doesn't discourage you from keeping on contributing! :rocket:

MadcapJake:
@noseglid thank you so much for your help! I thought that the `cmdDefaults` function was the deepest but didn't notice that there was an internal `createBuildConfig` function inside `settings`.

Perhaps that was the main reason why I was getting the generic timeout error messages. :relieved: 

Thanks!

noseglid:
Tested this, and it works well. But the branch has merge conflicts with master. Could you merge master to your branch and resolve conflicts?

noseglid:
Closing this due to inactivity.

PR #183, opened 2015-09-20
Fix example target JSON

Crocmagnon:
The JSON wasn't valid in the example target due to a missing comma.


noseglid:
Cool! Thanks for this!

PR #196, opened 2015-10-04
make targets available as atom commands

klorenz:
I like to have targets available as normal atom commands.  Having e.g. build command "GNU Make: default", you can access it from command palette as "Build GNU Make: default".

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -139,7 +139,11 @@ module.exports = {
       this.buildView.link(text, id, callback);
     });
 
-    this.refreshTargets().catch(function(e) {});
+    var _this = this;
+    var onInit = atom.packages.onDidActivateInitialPackages(function() {
+        _this.refreshTargets().catch(function(e) {});
+        onInit.dispose();
+    });
   },
 
   deactivate: function() {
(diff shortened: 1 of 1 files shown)


lib/build.js
```diff
@@ -139,7 +139,11 @@ module.exports = {
       this.buildView.link(text, id, callback);
     });
 
-    this.refreshTargets().catch(function(e) {});
+    var _this = this;
+    var onInit = atom.packages.onDidActivateInitialPackages(function() {
+        _this.refreshTargets().catch(function(e) {});
+        onInit.dispose();
+    });
```
noseglid: This uses babel, you should use `=>` instead to bind the context. Then you can use `this` in the function.

Also, for future reference, this is completely isolated from the rest of the changes you've made. Would've been good have this in a separate PR so that it could be merged regardless of the other code's status.

lib/build.js
```diff
@@ -236,19 +240,24 @@ module.exports = {
         });
 
         _.forEach(targets, function (target, targetName) {
-          if (!target.keymap) {
-            target.dispose = _.noop;
-            return;
+          var keymapDispose = null;
+          var commandName = 'build-' + targetName;
+          // there may be only one : in commandName
```
noseglid: Why can there only be one `:` ? It works well with multiple now.

I'm not sure why you need to replace anything at all from the command name.

noseglid:
closing due to inactivity. Please re-open if you intend to pursue this.

Issue #201, opened 2015-10-08
Error links in build panel are gone.


xpol:
There was once error links (with underline) in build panel. Click the link will  jump to source code line.

But recently, the Error links are gone, just plan text. 
*(The error parse is OK because the `build: error match` command works correctly.)*

noseglid:
Can you provide some output which reproduces this?

Is is possible to click the error and have it open the correct file still?

noseglid:
Hopefully fixed with #207. If not, open a new issue.

PR #207, opened 2015-10-11
[Bugfix] Fixes links not showing up

MackieLoeffel:
Fixes highlighting of some errors and maybe #201.

### DIFF
--- a/lib/build-view.js
+++ b/lib/build-view.js
@@ -230,7 +230,7 @@ module.exports = (function() {
   BuildView.prototype._render = function() {
     var string = _.escape(this.buffer.toString('utf8'));
     _.forEach(this.links, function (link, index) {
-      var replaceRegex = new RegExp(_.escapeRegExp(link.text), 'g');
+      var replaceRegex = new RegExp(_.escapeRegExp(_.escape(link.text)), 'g');
       string = string.replace(replaceRegex, '<a id="' + link.id + '">' + _.escape(link.text) + '</a>');
     });
     this.output.html(ansi_up.ansi_to_html(string));


noseglid:
What does the output look like that fixes this?

MackieLoeffel:
Something with characters, which are escaped by `_.escape`, like `<` or `>`
`at Object.<anonymous> (C:\\coding\\buildtest.coffee:2:10)`

noseglid:
So um... consider this output: `some stuff <match>`
Using the regex: `<match>` (if that was the errorMatch).

now it would be escaped so that it would look for `&lt;match&gt` in the string `some stuff <match>` which wouldn't hit.

Why doesn't it work as is?

MackieLoeffel:
The string, in which the links are inserted, is escaped as well beforehand. So the string is `some stuff &lt;match&gt;`
https://github.com/noseglid/atom-build/blob/ed6f0a5860cedc48875c36c86431ca217064fb72/lib/build-view.js#L231

you can't escape the string afterwards, because then you would escape the inserted `<a>` tags as well.

xpol:
I can confirm that this fixes #201.

Here is my errorMatch:

```json
  "errorMatch":[
    "\\[string \\\"(?<file>[^\\\"]+)\\\"\\]:(?<line>\\d+)",
    "line (?<line>\\d+) of chunk '\\\"(?<file>[^\\\"]+)\\\""
  ]
```

Maybe because of `[`, `]` or `"` in regex.

xpol:
Thanks @MackieLoeffel !

noseglid:
Cool. I see! Great fix @MackieLoeffel.

PR #210, opened 2015-10-20
better default font and match editor font

braver:
I think the output view looks a lot better when it uses the font set for the editor, instead of simply going with whatever the user agent default is for `monospace`. This updates the stylesheet to default to the same default fonts as the core stylesheet of Atom, and uses the configured font by creating an inline style on the output element.


noseglid:
Cool! Thanks for this!

braver:
:+1:

PR #211, opened 2015-10-22
Added general build file

C-Bouthoorn:
Placing a `.atom-build-defaults.json` file in `~/.atom/packages/build` makes it able to use general commands in all your projects.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -272,11 +272,11 @@ module.exports = {
     
     // Setting 'main' command
     
-    if( custom['name'] == undefined ) {
+    if ( custom['name'] == undefined ) {
       var key;
       
-       // Get first element from targets
-      for(key in targets) {
+      // Get first element from targets
+      for ( key in targets ) {
         break;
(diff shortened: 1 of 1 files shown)


C-Bouthoorn:
For example:

My `.atom-build-defaults.json` file looks like:

```json
{
  "Test2": {
    "cmd": "echo 'Test 2'"
  },
  "Test3": {
    "cmd": "echo 'Test 3'"
  }
}
```

which adds Test2 and Test3 to the general dropdown menu.

noseglid:
I don't think this is the way we want to go about and solve this.

If this should be part of this project, it should definitely be done in the atom-build tool (found in `lib/atom-build.js`). Maybe it would even be better to do this as a completely separate plugin to atom using [the service api](https://github.com/noseglid/atom-build#service-api-for-package-developers).

noseglid:
Closing due to inactivity. Please re-open if you intend to pursue this.

PR #212, opened 2015-10-22
[WIP] Linter integration.

kepler0:
I'm getting real close to having basic Linter integration.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -15,8 +15,6 @@ var GoogleAnalytics = require('./google-analytics');
 var ErrorMatcher = require('./error-matcher');
 var tools = [ require('./atom-build') ];
 
-var LinterIntegration = require('./linter-integration');
-
 function BuildError(name, message) {
   this.name = name;
   this.message = message;
@@ -488,7 +486,8 @@ export default {
   },
 
   consumeLinter: (linterInstance) => {
+    const LinterIntegration = require('./linter-integration');
     this.linterIntegration = new LinterIntegration(linterInstance);
-    console.log(this.linterInstance);
+    console.log(this.linterIntegration);
   }
 };
--- a/lib/build.js
+++ b/lib/build.js
@@ -15,6 +15,8 @@ var GoogleAnalytics = require('./google-analytics');
 var ErrorMatcher = require('./error-matcher');
 var tools = [ require('./atom-build') ];
 
+var LinterIntegration = require('./linter-integration');
+
 function BuildError(name, message) {
   this.name = name;
   this.message = message;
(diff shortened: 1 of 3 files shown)


lib/linter-integration.js
```diff
@@ -0,0 +1,31 @@
+'use babel';
+
+export default class LinterIntegration {
+  constructor(linterInstance) {
+    this.linter = linterInstance;
+    linterInstance.addLinter(this.provider);
+  }
+
+  get provider() {
```
steelbrain: This creates a new object everytime you access it, it'll confuse the base linter and result in duplicate messages. You should always use the **same** object.

kepler0: Thanks. This is the first time I've used ES6 classes, it felt a little derpy handling the object like that. I'll fix it.

steelbrain: You could just assign this same object to this instance in constructor

lib/linter-integration.js
```diff
@@ -0,0 +1,31 @@
...
+    return {
+      grammarScopes: [],
```
steelbrain: Missing a `name` field here.

lib/linter-integration.js
kepler0: Beautiful.

lib/linter-integration.js
kepler0: Name gets a little tricky for Build. I'll figure it out, eventually. I want to have the name be dynamically set for the actual build provider but that seems to be a little tricky. But I wasn't worried about it for this first basic implementation because the `name` field isn't required yet.

kepler0:
@steelbrain I think I'll wait for atom-community/linter#964 before I finalize this.

steelbrain:
Great. Here's my :+1:  on that

lib/build.js
```diff
@@ -372,6 +375,10 @@ module.exports = {
             if (atom.config.get('build.scrollOnError')) {
               this.errorMatcher.matchFirst();
             }
+            if (atom.packages.isPackageActive('linter')) {
```
noseglid: Might be better to check if `this.linterIntegration` is set here rather than the package activation, to avoid pesky race conditions.

noseglid:
I like what you've done here and look forward to the finalized thing.

noseglid:
Any progress on this?

kepler0:
There was an addition made to the Linter API to make this easier, I'll come back to this in a little bit.

steelbrain:
@kepler0 Just so you would know, I've just pushed a release with indie API, so you can continue working on this

kepler0:
@steelbrain Awesome. I'll work on this when I work a bit more on #223.

kepler0:
I'll post a new PR in a bit. I have some more work locally, but forgot about this open pull request and deleted the fork it was in.

PR #213, opened 2015-10-26
added management of relative paths within sub projects

patgithub:
I'm using qmake with several PRO files where usually sub projects are located within sub directories. Because of that the Makefiles end up in those sub directories but atom-build doesn't get that because CWD isn't changing.
The best solution would be to parse the "Entering directory X..." messages from make. But I'm not so familiar with js and so I found a more quick-hack solution.
Helps me a lot because now atom-build error messages lead me to the source code again.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -87,6 +87,13 @@ module.exports = {
       default: 'Bottom',
       enum: [ 'Bottom', 'Top', 'Left', 'Right' ],
       order: 8
+    },
+    subProjects: {
+      title: 'Sub Projects',
+      description: 'Comma separated list of sub folders or "*" for automatic (for example if you use qmake subdirs)',
+      type: 'string',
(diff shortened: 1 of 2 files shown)


oli-obk:
wouldn't it be better to modify your makefiles to produce absolute paths?

patgithub:
Yeah sure that would be wonderful, but until now I couldn't find a way how to tell this to qmake. Seems to me that qmake is not able to generate absolute paths.

noseglid:
I'm sorry but I do not want this complexity `atom-build` itself. I can think of two options at this point:
  * Fix the makefiles so that they output absolute paths (or at least relative to the cwd you specify).
  * Let build providers modify the output, and create a custom build provider for `qmake`.

The second obviously requires more work and modifications to the build provider interface.

Issue #198, opened 2015-10-06
Possible keybind overlap with Atom?


Jay1:
I don't do this very often, pardon my Github etiquette,

CTRL-ALT-B (build:trigger on atom-build) overlap with editor:move-to-previous-subword-boundary from atom-text-editor (which is the case by default I believe)

I don't know if it's a well known fact and I'm stating the obvious here, but I had to rebind build:trigger for a few months now, every new patch.

noseglid:
Which version of `build` do you have? This was already reported in #168 and (supposedly) fixed in #172 . This was released in `v0.46.0`.

Jay1:
version : 0.47.0

This is my build.json
```json
{
  ".platform-darwin atom-workspace, .platform-darwin atom-text-editor": {
    "cmd-alt-b": "build:trigger",
    "cmd-alt-v": "build:toggle-panel",
    "cmd-alt-g": "build:error-match",
    "cmd-alt-h": "build:error-match-first",
    "cmd-alt-t": "build:select-active-target"
  },
  ".platform-linux atom-workspace, .platform-linux atom-text-editor, .platform-win32 atom-workspace, .platform-win32 atom-text-editor": {
    "ctrl-alt-b": "build:trigger",
    "ctrl-alt-v": "build:toggle-panel",
    "ctrl-alt-g": "build:error-match",
    "ctrl-alt-h": "build:error-match-first",
    "ctrl-alt-t": "build:select-active-target"
  },
  ".build": {
    "escape": "build:stop"
  },
  ".build-confirm": {
    "enter": "build:confirm",
    "escape": "build:no-confirm"
  }
}

```

noseglid:
Yes, that is what the `build.json` keymap should look like. Could you bring up the Key-binding resolver (via the command palette) and press `ctrl-alt-b` or `cmd-alt-b` depending on OS? What does it show?

Jay1:
On windows 10
![Key-binding resolver](http://i.imgur.com/tAfRaE7.jpg)

Issue #168, opened 2015-08-25
Key binding is taken by built-in command


sryze:
Ctrl+Alt+B seems to be assigned to `editor:move-to-previous-subword-boundary` by default. Is there any way to choose one over the other?

noseglid:
I don't have that binding. Where does it come from?

You could set a keymap yourself, make sure it has the at least the same specificity as the built in. See https://discuss.atom.io/t/why-do-user-key-bindings-not-take-priority/11740

Reyncor:
Same issue after updating to 1.0.8
http://i.imgur.com/UTK0p81.png

That said... none of the F5-F10 keys are taken.
For now ill try to map them to

Function	Shortcut Key
Build	Ctrl + F9
Compile current file	Ctrl + Shift + F9
Run	Ctrl + F10
Build and Run	F9
Rebuild	Ctrl + F11

Which is codeblocks default and doesn't conflict with any Atom core shortcuts.

noseglid:
It appears as by default, `editor:move-to-previous-subword-boundary` is set to `alt-b` (See https://github.com/atom/atom/blob/master/keymaps/linux.cson#L98)

Is there something other that might have rebound the keys for you? Since you both have the same issue, it's likely to be another package. Would you mind pasting your installed packages (`apm list --installed`) ?

noseglid:
Hmm, I think I found it. It appears to be because of https://github.com/atom/atom/blob/master/keymaps/emacs.cson#L9.

This change seems to have been made recently in atom/atom. I've opened an issue to see the developers take on this.

Issue #163, opened 2015-08-21
Select Active Target hard to reach on Ubuntu: Ctrl-alt-T displays the terminal on all versions of Ubuntu.


Reyncor:
What can I do? For some reason atom build is not switching build targets when I have a file from another project open... always remembering the first project I built... not the one I'm on. I can't seem to switch the target. And ctrl+alt+T is a holy shortcut in Ubuntu which I love.

Is there a way to select the active target based on the file I have open?
Or at least change the shortcut on linux.

noseglid:
You can change the keybinding to whatever you want globally via the keymap settings:
Press `ctrl-alt-p`, type `open your keymap`. Add something like: 

```coffeescript
  "atom-workspace": 
    "ctrl-alt-t": "build:select-active-target"
```

And substitute `ctrl-alt-t` with whatever you want.

You could also assign a custom keymap to separate builds in the custom build file configuration, key `keymap`.

Reyncor:
Perfect... made it be Ctrl-Alt-Shift-B

Still think it should be the default on at least Linux.

Thank you! ( btw you meant "Press ctrl-shift-p" )

Reyncor:
Almost forgot.... This wouldn't even be necessary if the Active Target would auto switch based on which project the currently highlighted file was from. Is there a way to do that?

noseglid:
Do you mean when it's using multiple project folders in one Atom window? If so, I've been meaning to change the behavior there so that it will remember which target was selected per project, and if any file in that project is currently active, it will run that target. Right now it's kind of silly that it builds `Project A` even if you're currently working in `Project B` just because you happened to build that last.

Reyncor:
Exactly my point... Though pressing a 4 key shortcut isn't horrible for now... Thank you.

PR #215, opened 2015-10-29
Improve default keybindings for Linux/Windows

jhasse:
Currently the default key bindings on Linux and Windows aren't optimal. For example `ctrl-alt-t` is already used by most distros to open a terminal, see #163. In general it's uncommon for Linux apps to use `ctrl-alt-*` keybindings as they are usually used by system apps.

After looking through the issues, it looks to me that only people on Linux/Windows have problems with the default keybindings: #168, #198.

Also note that `ctrl-alt-*` has some problems on Windows, see this post which advises not to use them on Windows at all: https://github.com/atom/atom-keymap/issues/35#issuecomment-117284074

Last but not least they are rather uncommon among existing build apps like Sublime Text or CodeBlocks.

I've tried to find some new defaults, here is my reasoning for them:

* `f4` error-match: This is the keybindings Sublime Text uses.
* `shift-f4` error-match-first: Should be similar to error-match, therefore added shift.
* `f9` trigger: This is used by CodeBlocks for build and run
* `f7` select-active-target: To match Visual Studio's build shortcut. Maybe change to `shift-f9`?
* `f8` toggle-panel: A key between f7 and f9 :D I'm not sure about this one.

Also see https://github.com/noseglid/atom-build/issues/168#issuecomment-136391472 for suggestions.

What do you think? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 
 Automatically build your project inside your new favorite editor, Atom.
 
-  * `cmd-alt-b` / `ctrl-alt-b` builds your project.
-  * `cmd-alt-g` / `ctrl-alt-g` cycles through causes of build error. See [Error Matching](#error-match).
-  * `cmd-alt-h` / `ctrl-alt-h` goes to the first build error. See [Error Matching](#error-match).
-  * `cmd-alt-v` / `ctrl-alt-v` Toggles the build panel.
-  * `cmd-alt-t` / `ctrl-alt-t` Displays the available build targets.
+  * `cmd-alt-b` / `ctrl-alt-b` / `f9` builds your project.
+  * `cmd-alt-g` / `ctrl-alt-g` / `f4` cycles through causes of build error. See [Error Matching](#error-match).
+  * `cmd-alt-h` / `ctrl-alt-h` / `shift-f4` goes to the first build error. See [Error Matching](#error-match).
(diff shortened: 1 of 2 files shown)


noseglid:
Thanks for looking into this so thoroughly! I like your reasoning and I'd be happy to merge changes to the keybinding. However, I think we will be better off if we leave the old keybindings in place and just add the keybindings you're suggesting. I don't see any reason to restrict them to Windows/Linux either.

jhasse:
You're right, that will result in fewer surprises when users update ;)
How can I add multiple keymaps?
Should all three be documented?

noseglid:
You should add something like this to the `keymaps.json`
```
"atom-workspace, atom-text-editor": {
  "f9": "build:trigger",
  "f8": "build:toggle-panel",
  "f4": "build:error-match",
  "shift-f4": "build:error-match-first",
  "f7": "build:select-active-target"
}
```
in addition to what's already there.

I think it's sane to document them all.

jhasse:
Done :)

In the text, I left the changes for recommending `cmd-alt-*` for OS X and the new keybindings for Linux/Windows. IIRC f keys are rarely used on Mac and always need the Fn key on Macbooks.

noseglid:
Great. thanks!

PR #218, opened 2015-11-03
Use regexps with '/g' modifier to replace all matches of built-in vars.

nsf:
As far as I understand javascript's "replace" with string arguments does only single replacement. I needed to replace all matches when building some of the Go projects. The use case was to redefine an environment variable like so: `"GOPATH": "{PROJECT_PATH}:{PROJECT_PATH}/vendor"`. With this fix it actually works.


noseglid:
Cool. Thanks!

PR #219, opened 2015-11-04
improve layout in sidebar mode

braver:
Some css tweaks:

- use `@component-padding` from ui-theme for padding instead of `10px`
- set minimum width
- all buttons in a row so output gets to use the entire width of the panel

Before  
before

After
after


noseglid:
Very nice. Thank you for this!

PR #223, opened 2015-11-12
[WIP] Optimize for ECMAScript 6.

kepler0:


lib/google-analytics.js
```diff
@@ -52,30 +44,35 @@ module.exports = (function () {
 
     GoogleAnalytics.getCid(function (cid) {
       _.extend(params, { cid: cid }, GoogleAnalytics.defaultParams());
-      GoogleAnalytics.request('https://www.google-analytics.com/collect?' + querystring.stringify(params));
+      this.request('https://www.google-analytics.com/collect?' + querystring.stringify(params));
```
noseglid: `this` is not defined here. You'd need to change the function definition to `(cid) => {`.

kepler0: Good catch. Thanks. I've been using the unit tests to make sure that nothing breaks. If it's not covered by unit tests, I won't catch it.

steelbrain:
While you are doing this, may I suggest [etch](https://github.com/nathansobo/etch) instead of space-pen-views?

kepler0:
@steelbrain Different Issue, maybe? I just want to patch this so it is using ECMAScript 2015 classes.

steelbrain:
You have my :+1:

kepler0:
woot

lib/google-analytics.js
noseglid: Google analytics are left out of specs as of now. Not sure if it make sense to test it or not. If it breaks it wont affect user experience anyway

kepler0: I'll end up testing this in my own environment when I'm done, so any glaring issues should be taken care of then.

noseglid:
This is finalized in #229

kepler0:
Awesome!

PR #232, opened 2015-11-22
Fixed dependency.

C-Bouthoorn:
`package.json` uses `xregexp` as package name, but you use `XRegExp` in your project. Fixed in this file.

### DIFF
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -2,7 +2,7 @@
 
 import fs from 'fs';
 import path from 'path';
-import { XRegExp } from 'XRegExp';
+import { XRegExp } from 'xregexp';
 import { EventEmitter } from 'events';
 import GoogleAnalytics from './google-analytics';
 


noseglid:
That is definitely a regression. Thanks for spotting and fixing this. I'll release a patch with this!

Reason I didn't notice it is because I run on a case-insensitive file system (OS X with HFS, sigh..). Your fix is needed for it to work on a case-sensitive.

kepler0:
@noseglid That would explain why neither of us found this error.

kepler0:
I'm curious why it passed unit tests on Travis though

noseglid:
I think travis has OS X hosts too, and it defaults to those. The [travis output](https://travis-ci.org/noseglid/atom-build/builds/92535674) says
```
Build system information
Build language: objective-c
Build group: stable
Build dist: osx
```

kepler0:
Ah... Right. I forgot the Travis does Atom packages on OS X by default.

C-Bouthoorn:
@noseglid Might be worth it to change configs in case you mess with capitalization later on. Or just wait for people like me running Linux to test it :tongue:

noseglid:
On my todo list is to run them on both OSX and Linux, as well as windows using appveyor. But there's only so much time. Any contributions for this is greatly appreciated.

Issue #216, opened 2015-10-29
:/usr/local/bin added to Windows path when plugin is enabled


dabowheel:
:/usr/local/bin is added by my windows path when this plugin is enabled. This causes commands not to be found within Atom when they should. You can see the path by getting it in the console accessed by Ctrl-Alt-i.

process.env.path
"C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;C:\ProgramData\Oracle\Java\javapath;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files (x86)\Brackets\command;C:\Program Files\PostgreSQL\9.4\bin;C:\Python27;C:\Program Files (x86)\PostgreSQL\9.4\bin;C:\php;C:\Program Files (x86)\Git\cmd;C:\Program Files (x86)\Heroku\bin;C:\Program Files (x86)\git\cmd;C:\ProgramData\ComposerSetup\bin;C:\Program Files\nodejs\;C:\Program Files (x86)\Calibre2\;C:\Program Files\MongoDB\Server\3.0\bin;C:\Users\Miguel\AppData\Local\atom\bin;C:\Users\Miguel\AppData\Roaming\npm:/usr/local/bin"

dabowheel:
The problem is mainly assuming that the path delimiter is ":" when it can be ";" on Windows. Using path.delimiter in the Node path module instead of hard coding the delimiter would fix this.

dabowheel:
Place in code where this happens:
$ grep "/usr/local/bin" . -r
./lib/build.js:    // Manually append /usr/local/bin as it may not be set on some systems,
./lib/build.js:    process.env.PATH += ':/usr/local/bin';

noseglid:
That's a legacy thing from before Atom was supported on windows. It should definitely be removed (or at least only appended when running on Linux/OSX. Thanks for reporting!

PR #233, opened 2015-11-22
Remove PATH edit from Package Activation.

kepler0:
This can cause issues on non *nix platforms and it's outside of build's scope.
Users with environment issues should install atom-community's [`environment` package](https://atom.io/packages/environment).

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -80,11 +80,6 @@ module.exports = {
   },
 
   activate: function () {
-    // Manually append /usr/local/bin as it may not be set on some systems,
-    // and it's common to have node installed here. Keep it at end so it won't
-    // accidentially override any other node installation
-    process.env.PATH += ':/usr/local/bin';
-
     this.buildView = new BuildView();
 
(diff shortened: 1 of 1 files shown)


kepler0:
Fixes #216

noseglid:
As we talked about on slack; Should be changed to append only for unix like systems rather than removed completely, so we don't break functionality for people relying on this right now.

noseglid:
Yes that should be it. os is a [module from node](https://nodejs.org/api/os.html).

C-Bouthoorn:
@noseglid So `os=require('os');` is correct in this case? As I see `include ... from ...` everywhere.

noseglid:
@C-Bouthoorn it'd be better to to `import os from 'os';` to be more ES6y

C-Bouthoorn:
@noseglid Ah, okay.
@kepler0 Would you please implement this code, so noseglid can pull? :smile:

noseglid:
This was fixed in #247

PR #240, opened 2015-11-25
Newer CSON-parser

noseglid:
Also added specs for CSON files


noseglid:
I will bring this in since it's an improvement (also some useful specs). We will still need to do more work to support regexes in the CSON file (As reported in #246 )

Issue #224, opened 2015-11-14
matched errors should not change the build status


dirk-thomas:
If the invoked build tool returns with a code of 0 the error matching should not change the build result to *failed*. If this doesn't seem reasonable as a default there should be an option to switch the behavior.

noseglid:
This was actually added in #144 . I suppose an option to control this is in order.

dirk-thomas:
I commonly trust by build tools return code more then the regular expression which might match if the output e.g. "by accident" contains a certain string. It would be great there would be an option for this to choose the desired behavior.

noseglid:
fixed in #224

PR #244, opened 2015-11-30
add option to toggle if build should fail if error regex matches

dirk-thomas:
This PR addresses #224. It adds a new option to allow disabling the behavior added in #144. The default value does not change the current behavior.

I added the new option `matchedErrorFailsBuild` after the build trigger options and before the option determining the behavior after the build (scrollOnError).

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -37,19 +37,26 @@ export default {
       default: false,
       order: 3
     },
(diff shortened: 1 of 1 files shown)


noseglid:
Very nice. Thank you!

PR #251, opened 2015-12-05
Linter integration

keplersj:
Currently only providing basic integration, little buggy. But it does work!


kepler0:
@steelbrain, thoughts?

lib/build.js
```diff
@@ -338,6 +343,11 @@ export default {
     targetName = targetName || this.activeTarget[p];
 
     Promise.resolve(this.targets[p]).then(targets => {
+      if (atom.packages.isPackageActive('linter')) {
```
steelbrain: `if (this.linterRegistry !== null)`

kepler0: Of course, `linter` may not be the only package providing the linterRegistry. Just the most likely...

steelbrain: It's not about giving `linterRegistry`, it's about minimizing usage of external API

noseglid: Would it make sense to do this in the callback where all packages are activated (this already exists in `activate`) ?

kepler0:
@steelbrain @noseglid I just push a commit to test against the linter registry variable, rather than linter itself. Anything else?

noseglid:
I get this error:
```
Error: Missing html/text field on Linter Response
    at /Users/noseglid/.atom/packages/linter/lib/validate.coffee:40:19
    at Array.forEach (native)
    at Object.module.exports.Validate.messages (/Users/noseglid/.atom/packages/linter/lib/validate.coffee:27:14)
    at MessageRegistry.set (/Users/noseglid/.atom/packages/linter/lib/message-registry.js:36:16)
    at /Users/noseglid/.atom/packages/linter/lib/linter.coffee:35:17
    at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:86:11)
    at /Users/noseglid/.atom/packages/linter/lib/indie-registry.js:27:20
    at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:86:11)
    at Indie.setMessages (/Users/noseglid/.atom/packages/linter/lib/indie.js:15:18)
    at ChildProcess.<anonymous> (/Users/noseglid/devel/atom-build/lib/build.js:423:23)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:172:7)
    at maybeClose (internal/child_process.js:817:16)
    at Socket.<anonymous> (internal/child_process.js:319:11)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at Pipe._onclose (net.js:469:12)
```

I think it is because `error.message` is not defined. Please see my inline comments

lib/build.js
```diff
@@ -408,6 +418,20 @@ export default {
...
+          this.linter.setMessages([{
+            type: error.type || 'Error',
+            name: targetName,
+            text: error.message,
+            filePath: error.file,
+            range: [
+              [error.lineStart - 1 || error.line - 1,
+                error.columnStart - 1 || error.column - 1],
+              [error.lineEnd - 1 || error.line - 1,
+                error.columnEnd - 1 || error.column - 1]
```
noseglid: error set `col`, not `column`.

`lineStart`, `columnStart`, `lineEnd` and `columnEnd` is never set on `error` as far as I can tell. Why have these?

kepler0: I'm attempting to use both the standard `build` named captures, and `linter` named captures.

kepler0: I'll add a `col` condition when I sit back down.

kepler0:
@noseglid What provider were you testing with? It was working with `build-xcodebuild`

noseglid:
Also, when a build is triggered anew, it should clear the linter error.

kepler0:
I'll take care of clearing as well.

noseglid:
I was testing with this:
```
{
  "cmd": "echo hello world: .eslintrc:2:13",
  "errorMatch": [
    "hello world: (?<file>[^:]+):(?<line>\\d+):(?<col>\\d+)"
  ]
}
```
Make sure you have `Any matched error will fail the build` option checked out in atom-build settings

kepler0:
Hmm... I'll have to add a case for build errors that don't supply error
text.

noseglid:
`message` is not part of the [documentation for error matching](https://github.com/noseglid/atom-build#error-matching). I'd venture to say most error matchers don't have `message` set.

It seems as the error matcher in `build-xcodebuild` doesn't follow the documentation.

I would be open to adding `message` to the documentation for the benefit of the linter integration. There may very well be a use for it outside of that as well.

noseglid:
We must also take into consideration that only `file` is actually required. neither line nor col is a must. The error matcher will be as specific as provided.

lib/build.js
```diff
@@ -408,6 +418,20 @@ export default {
           }
           GoogleAnalytics.sendEvent('build', 'failed');
         }
+        for (const error of this.errorMatcher.currentMatch) {
+          this.linter.setMessages([{
+            type: error.type || 'Error',
+            name: targetName,
```
steelbrain: I'm adding this message-specific name to linter now. It'll be in the next release.

kepler0: Awesome.

oli-obk:
any updates on this?

noseglid:
Closing this in favor of #354.

PR #255, opened 2015-12-11
(partly) cross-platform build specs via cmd (sh:true) for windows

dreamair:
(partly) because arguments might still depend on the platform (windows path separators). Since this is less an issue with node processes, this enhancement can still be valuable.


noseglid:
You gotta fix [the lint errors](https://travis-ci.org/noseglid/atom-build/builds/96312227) travis complains about.

I think this looks nice, but I'm not much of a windows user.

C-Bouthoorn:
Should be fixed in https://github.com/dreamair/atom-build/pull/2

noseglid:
Thanks!

PR #260, opened 2015-12-16
Add Leftmost and Rightmost settings for status bar position.

xpol:
* Add `Leftmost` and `Rightmost` position settings for statusbar item.
* Calculated priority rather than hard coded one. 

*Note: `Leftmost` behaviour as origin `Left`.*

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -99,8 +99,8 @@ export default {
       title: 'Status Bar',
       description: 'Where to place the status bar. Set to `Disable` to disable status bar display.',
       type: 'string',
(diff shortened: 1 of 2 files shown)


xpol:
@noseglid 
Will you please take a look at PR.
Thank you very much!

noseglid:
I'm thinking - would it make more sense to be able to insert the priority directly? That way it can optionally be put wherever on the left or right side of the status bar.

xpol:
You are right.
Let user search/try the desired priority.

Issue #209, opened 2015-10-13
Global build targets


tmazur:
As far as I am aware, there is no way to define build targets that are available across all projects (even those without .atom-build files). Providing such a feature would be a nice enhancement.

academo:
I can related to this, is any way to include a global config file beside the hard coded plugins?

noseglid:
There is no way right now. But I'm looking into adding additional paths to search for the custom files (`.atom-build.json`, `.atom-build.cson` - and soon `.atom-build.yml`).

The user home directory would be an obvious candidate. Possibly also a configurable path to search for global build targets.

academo:
@noseglid home path makes a lot of sense if you think about other plugins doing that, also, what about those projects with multiple folders from different paths (my case for instance)?

C-Bouthoorn:
I also tried this in https://github.com/noseglid/atom-build/pull/211 , but that didn't work out well, as I created a new file containing both files. Currently I'm trying to do something with [atom-build-defaults](https://github.com/C-Bouthoorn/atom-build-defaults) (Already made files earlier, just pushed them), but it isn't working yet. I'll see what I can do with that, and will report back!

academo:
@noseglid I haven't read the whole code but I see in `lib/atom-build.js` line 53 there is the `isEligible()` method, which is looking for the files to be parsed and it is reducing the scope of search to the specific cwd, if is easy as creating an alternative cwd with the user home path I can make a pull request when I get the moment to code the change

C-Bouthoorn:
@academo Yeah, but *afaik* it tries to merge them, and that can have some weird effects when they have the same name for example. I've already tried such things on my own, and they worked until you actually added a `.atom-build.json` to the project, then it crashed :wink:

academo:
@C-Bouthoorn maybe reduce the scope, first find the closest inside the project and fallback to the home directory.

C-Bouthoorn:
@academo That's an option. Let me try something, and I will come back to it

C-Bouthoorn:
```javascript
  isEligible() {
    this.files = [ '.atom-build.json', '.atom-build.cson' ]
      .map(file => path.join(this.cwd, file))
      .filter(fs.existsSync);
      
    if( 0 < this.files.length) {
      const atom = path.join(os.homedir(), '.atom', 'packages', 'build');
      
      this.files = [ '.atom-build.json' '.atom-build.cson' ]
        .map(file => path.join(atom, file))
[...]

PR #264, opened 2015-12-18
Add posibility to use an atom-build config file from the user's home

academo:
As discussed on https://github.com/noseglid/atom-build/issues/209

Adding the possibility to use global config files, as a first step, to use the user's home directory looking for any of the supported config file types.

as a bug: the plugin can't watch the whole directory structure, if an atom build file is added to the home directory after the project is loaded, atom build won't recognize the new file, the project needs to be reloaded.

This commit may looks big but tests required lots of modification to support the new "isElegible" method fallback to use user home directory.

### DIFF
--- a/spec/build-confirm-spec.js
+++ b/spec/build-confirm-spec.js
@@ -4,16 +4,20 @@ import _ from 'lodash';
 import fs from 'fs-extra';
 import temp from 'temp';
 import specHelpers from 'atom-build-spec-helpers';
+import os from 'os';
 
 describe('Confirm', () => {
   let directory = null;
   let workspaceElement = null;
   const cat = process.platform === 'win32' ? 'type' : 'cat';
   const waitTime = process.env.CI ? 2400 : 200;
+  const originalHomedirFn = os.homedir;
 
   temp.track();
 
   beforeEach(() => {
+    const createdHomeDir = temp.mkdirSync('atom-build-spec-home');
+    os.homedir = () => createdHomeDir;
     directory = fs.realpathSync(temp.mkdirSync({ prefix: 'atom-build-spec-' })) + '/';
     atom.project.setPaths([ directory ]);
 
@@ -37,6 +41,7 @@ describe('Confirm', () => {
 
   afterEach(() => {
     fs.removeSync(directory);
+    os.homedir = originalHomedirFn;
   });
 
   describe('when the text editor is modified', () => {
--- a/spec/build-spec.js
+++ b/spec/build-spec.js
@@ -4,6 +4,7 @@ import fs from 'fs-extra';
 import path from 'path';
 import temp from 'temp';
 import specHelpers from 'atom-build-spec-helpers';
+import os from 'os';
 
 describe('Build', () => {
   const goodAtomBuildfile = __dirname + '/fixture/.atom-build.json';
@@ -13,6 +14,7 @@ describe('Build', () => {
   const shTrueAtomBuildFile = __dirname + '/fixture/.atom-build.sh-true.json';
   const shDefaultAtomBuildFile = __dirname + '/fixture/.atom-build.sh-default.json';
   const syntaxErrorAtomBuildFile = __dirname + '/fixture/.atom-build.syntax-error.json';
+  const originalHomedirFn = os.homedir;
 
   let directory = null;
   let workspaceElement = null;
@@ -41,6 +43,11 @@ describe('Build', () => {
       }).then( (dir) => {
         directory = dir + path.sep;
         atom.project.setPaths([ directory ]);
+        return specHelpers.vouch(temp.mkdir, 'atom-build-spec-home');
+      }).then( (dir) => {
+        return specHelpers.vouch(fs.realpath, dir);
+      }).then( (dir) => {
+        os.homedir = () => dir;
         return atom.packages.activatePackage('build');
       });
     });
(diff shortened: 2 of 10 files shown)


academo:
Checking the issues and pushing a fix...

C-Bouthoorn:
Fixed linter issues: https://github.com/academo/atom-build/pull/1

academo:
OK, last commit finally fixing the eslint issues, I think I had to change the lint plugin from the beginning, I am using jshint.

spec/build-spec.js
```diff
@@ -35,6 +37,8 @@ describe('Build', () => {
         return specHelpers.vouch(fs.realpath, dir);
       }).then( (dir) => {
         directory = dir + '/';
+        const createdHomeDir = temp.mkdirSync('atom-build-spec-home');
+        os.homedir = () => createdHomeDir;
```
noseglid: I think this should be made async just as it is done a few lines up, also - run it through `fs.realpath` if the temporary directory is a symlink

spec/custom-provider-spec.js
```diff
@@ -3,23 +3,48 @@
...
+    });
+    it('should find cson file in home directory', () => {
+      fs.writeFileSync(createdHomeDir + '/.atom-build.cson', fs.readFileSync(__dirname + '/fixture/.atom-build.cson'));
+      expect(builder.isEligible()).toEqual(true);
+    });
+    it('should find yml file in home directory', () => {
+      fs.writeFileSync(createdHomeDir + '/.atom-build.yml', fs.readFileSync(__dirname + '/fixture/.atom-build.yml'));
+      expect(builder.isEligible()).toEqual(true);
+    });
+  });
```
noseglid: It would also be cool with a test to make sure the targets are pulled out of the global file correctly.

spec/build-confirm-spec.js
```diff
@@ -4,14 +4,19 @@ import _ from 'lodash';
 import fs from 'fs-extra';
 import temp from 'temp';
 import specHelpers from 'atom-build-spec-helpers';
+import os from 'os';
+
```
noseglid: One new line too many (I'm suprised the linter doesn't complain here)

spec/build-spec.js
academo: only in this spec is using async, in all the others specs is being done sync way.

is it really necessary for a test?

spec/custom-provider-spec.js
academo: ++ I will add this

spec/build-confirm-spec.js
academo: I am not sure I understand what you mean here.

spec/build-spec.js
noseglid: Since the working directory is generated async, it makes sense to keep it consistent.
Should be easy enough to copy from that.

academo: Yes, I was going to do it like that but then I saw all the others tests doing sync, I will change that

spec/build-confirm-spec.js
noseglid: It's just that row 8 is unnecessary and there are now two blank lines between the import statement and the next statement: http://eslint.org/docs/rules/no-multiple-empty-lines

The ESlint configs for atom-build [requires this](https://github.com/AtomBuild/eslint-config-atom-build/blob/master/stylistic-issues.js#L64), but that was added in a later version of that config. I will upgrade, but until then, feel free to fix or ignore this comment as you please.

academo: I see, rather fix, I will do

spec/custom-provider-spec.js
academo: @noseglid I rethought this and I think it is not necessary, since this tests are about being able to read the file from home directory and it does it, It is not about parsing targets.

That being said, I think the tests after mine it handles the parsing and checking the files are being parsed.

I would suggest we better split the tests logic into: 1, check it is able to find a config file from the directory/project, 2: check it is able to parse the config file and the parsed data is correct.

Otherwise we would be me mixing two differents test into one.

I updated the PR with the async logic for home directory, realpath checking and the extra line in the os import.

Please advise and looking forward to keep contributing if you want!

noseglid: Absolutely. A separate test to make sure it parses targets. I don't think it is necessary to test all options and file variations, just one file and make sure it gets the target from there should do it.

This looks great and works well from my manual testing! I'd be happy to merge this when you add a test like that!

noseglid:
Any progress on this?

academo:
@noseglid crazy of a week + holidays = no time yet, I have it in my TODO list.

noseglid:
Excellent. There's no rush. Just wanted to keep myself updated :)

noseglid:
No cigar on this one?

I think this can be quite useful to have. If you don't intend to continue, let me know and I'll try to pick it up, or someone else may

academo:
@noseglid I am giving a try right now, It looks like now is full of conflicts

academo:
@noseglid fixed conflicts and updated tests again:

Note in the custom provider I splitted the tests in 3 parts: Check targets in home directory, check targets in project directory and parse/check parsed targets.

As it modified too many files is easy to get a conflict with other chaanges.

noseglid:
Really nice. Thank you for this!

PR #266, opened 2015-12-19
Initialize undefined environment variable properly

akfish:
Fields of `process.env` could be `undefined` in some environment (for example, mine):
image

This will cause the following error:
err

It can be fixed by simply initializing `replace`'s `value` parameter with empty string by default.


noseglid:
Thanks you!

PR #280, opened 2016-01-11
Changes editor font size will also affect build view.

xpol:


noseglid:
Splendid. Thanks!

xpol:
Thank you very much!

Issue #277, opened 2016-01-08
Crash when saving .atom-build.json


cwaffles:
Hi,

Whenever I save my .atom-build.json file, it crashes/freezes the atom editor. Sometimes I don't get the warning dialogue, but it shows up 9/10 times. The actual builds work fine, since the file is saved before it crashes.

![screenshot at 2016-01-08 14 02 58](https://cloud.githubusercontent.com/assets/3838856/12210688/a834f548-b610-11e5-9039-85270b8e0398.png)

System: Linux Mint 17.3 x64 MATE
Atom version 1.3.2
All community plugins disabled
Config file: http://pastebin.com/DqsuBDpn

C-Bouthoorn:
Confirmed in Linux 4.2.0-23 Ubuntu 64 bit   Atom 1.4.0-beta2

noseglid:
That is weird. I guess it has something to do with the filewatcher being used which is supposed to update the targets when the file is changed. Does it work if it is edited with another editor, or does it crash then as well? It should be enough to just `touch` the file on linux.

cwaffles:
When editing the .atom-build.json file in another editor, there is no crash, and the build plugin gives me the blue notification in the top right corner of my screen that the targets are found.

noseglid:
I see. This doesn't seem to be reproducible on OS X (which I mainly work on). Did you try reporting to `atom/atom` as suggested?

C-Bouthoorn:
Okay, so when I change the .atom-build.json file in another editor (nano), Atom _does_ crash.

When running in dev mode (`atom-beta -d -f -w`), I get

`[14971:0100/000000:WARNING:ipc_channel_posix.cc(824)] Message needs unreceived descriptors channel:0x2119ddc9b600 message-type:1704264 header()->num_fds:1`

Editing the file with Build disabled gives no message, and Atom doesn't crash.

C-Bouthoorn:
Update:

Changing `() => this.emit('refresh')` to `() => console.log("Hello")` on line 62 in atom-build.js stopped the error from coming. Doing further checks.

noseglid:
Thank you for investiaging. Using `fs.watchFile` works on OS X. However, I suspect there may be some race condition going on here which probably is not fixed by this change. I guess the weird thing is that it `fs.watch` seem to trigger even though the file is only altered once.

[...]

PR #281, opened 2016-01-11
Update atom-build.js

C-Bouthoorn:
Fixed problem where Atom would crash on Linux.

As discussed [here](https://github.com/noseglid/atom-build/issues/277)

### DIFF
--- a/lib/atom-build.js
+++ b/lib/atom-build.js
@@ -59,7 +59,7 @@ export default class CustomFile extends EventEmitter {
 
   settings() {
     this.fileWatchers.forEach(fw => fw.removeAllListeners());
-    this.fileWatchers = this.files.map(file => fs.watchFile(file, () => this.emit('refresh')));
+    this.fileWatchers = this.files.map(file => fs.watchFile(file, { persistent: false }, () => this.emit('refresh')));
 
     const config = [];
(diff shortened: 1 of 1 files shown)


noseglid:
It appears it returns a `StatWatcher` instead, which does not have a close function. It is an event emitter so you must call `removeAllListeners()` instead.

noseglid:
@C-Bouthoorn ping on this :)

C-Bouthoorn:
Oh, whoops, good that you ping. Will fix!

C-Bouthoorn:
Okay, so Travis doesn't like it, but is this problem related to this?

noseglid:
No I think they are related. One is about cached content of atom-build. Which the file watchers are in charge of not doing. I tried re-triggering the test and it fails with same error. I also get those errors locally.

C-Bouthoorn:
Okay, so it's clear that it doesn't work. However, I have no idea how I should fix this. It looks like `.watchFile()` just caches the file, and that that can't be changed. I don't have any idea why the timeout happens either. I'm sorry

noseglid:
Too bad.. this was circumstantial at best anyway. I think we should attempt to get to the root cause of the crash rather than this work around. I'll close this PR.

PR #295, opened 2016-01-23
Register target commands regardless of keymap existence

orgkhnargh:
Commands like `build:trigger:<command name>` were previously registered
only in case when command definition contains keymap.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -3,7 +3,7 @@
 import fs from 'fs';
 import path from 'path';
 import _ from 'lodash';
-import { Disposable } from 'atom';
+import { Disposable, CompositeDisposable } from 'atom';
 import kill from 'tree-kill';
 import Grim from 'grim';
 
@@ -210,20 +210,21 @@ export default {
         this.targets[p].forEach(target => target.dispose());
 
         settings.forEach((setting, index) => {
-          if (!setting.keymap) {
-            return;
+          const subscriptions = new CompositeDisposable();
+
+          const commandName = 'build:trigger:' + setting.name;
+          subscriptions.add(atom.commands.add(
+            'atom-workspace', commandName, this.build.bind(this, 'trigger')
+          ));
+
+          if (setting.keymap) {
+            GoogleAnalytics.sendEvent('keymap', 'registered', setting.keymap);
+            const keymapSpec = { 'atom-workspace, atom-text-editor': {} };
+            keymapSpec['atom-workspace, atom-text-editor'][setting.keymap] = commandName;
+            subscriptions.add(atom.keymaps.add(setting.name, keymapSpec));
           }
 
-          GoogleAnalytics.sendEvent('keymap', 'registered', setting.keymap);
-          const commandName = 'build:trigger:' + setting.name;
-          const keymapSpec = { 'atom-workspace, atom-text-editor': {} };
-          keymapSpec['atom-workspace, atom-text-editor'][setting.keymap] = commandName;
-          const keymapDispose = atom.keymaps.add(setting.name, keymapSpec);
-          const commandDispose = atom.commands.add('atom-workspace', commandName, this.build.bind(this, 'trigger'));
-          settings[index].dispose = () => {
-            keymapDispose.dispose();
-            commandDispose.dispose();
-          };
+          settings[index].dispose = () => subscriptions.dispose();
         });
 
         this.targets[p] = settings;


noseglid:
The names of the targets in the command palette are quire similar though:
screen shot 2016-01-24 at 17 58 31

According to the atom documentation:

> Command names must follow the namespace:action pattern, where namespace will typically be the name of your package, and action describes the behavior of your command. If either part consists of multiple words, these must be separated by hyphens. E.g. awesome-package:turn-it-up-to-eleven. All words should be lowercased.

I am aware that this is not new in this PR, but arguably quite more noticible.

One solution could be to `string.replace(/[^a-z\-]/g, '')`. The same would have to be done at row `325` where we find the target by name.

Do you have another suggestion?

orgkhnargh:
Your solution will make command names follow the required pattern, but it will also make them obscure. Here is the example:

```js
> 'PlatformIO: Build'.replace(/[^a-z\-]/g, '')
< "latformuild"
```

My suggestion is to do the following:

```js
function normalizeCommandName(commandName) {
  return commandName.toLowerCase()
    // Commands usually have their own namespace. 
    // This will remove a space between namespace and the actual command.
    .replace(": ", ":")
    // Replace whitespace with hyphen
    .replace(/\s+/g, "-")
}
```

Examples:

```js
> normalizeCommandName('PlatformIO: Build')
< "platformio:build"
> normalizeCommandName('Prefix: Four Words Command Name')
< "prefix:four-words-command-name"
```

noseglid:
`toLowerCase()` is a good idea.
I do not think we can have the colon there though. it appears that Atom takes the name of the command (e.g. `a:b:c`) splits on colon and takes the first entry (index 0) and second entry to build the commandname - which mean `c` would be removed in my example.

How about `toLowerCase()` then the regex I suggested?

orgkhnargh:
I just checked it out myself. It appears that atom does not perform any manipulation on command name, Atom saves it as is. Run this into developer console:

```js
atom.commands.findCommands({target: atom.views.getView(atom.workspace) })
  .map(function(command) {return command.name})
  .filter(function(name) {return name.indexOf("build") === 0})
```

On Atom 1.4.0 I can see this as a result:

```js
["build:toggle-panel", "build:error-match", "build:error-match-first", "build:refresh-targets", "build:trigger", "build:select-active-target", "build:stop", "build:confirm", "build:no-confirm", "build:trigger:PlatformIO: Build", "build:trigger:PlatformIO: Clean", "build:trigger:PlatformIO: Upload"]
```

I think that `namespace:action` is only intended to decrease the probability of name conflicts. I do not see any reason, why `action` cannot have its own namespace as well.

Also, I think that provider writers may be willing to assign the command name explicitly. So how about adding optional `commandName` parameter to target definition, e.g.:

```json
{
  "cmd": "<command to execute>",
  "name": "<name of target>",
  "commandName": "<name of Atom command>",
  "args": [ "<argument1>", "<argument2>" ],
}
```

But fallback to what you've suggested:
```js
.map(tool => {
  if (!tool.commandName) {
    tool.commandName = tool.commandName.toLowerCase().replace(/[^a-z\-]/g, '');
  }
  return tool;
})
```
under [this line](https://github.com/noseglid/atom-build/blob/master/lib/build.js#L179). Then full command might be like `build:trigger-{commandName}`.

noseglid:
It really appears to register with colons and stuff. I guess it's mainly presentation in the command palette which is filtered.

Supplying an additional key is certainly an option. To further distinguish it from `cmd` maybe we should call it `atomCommandName` instead? I think we should throw an error if the commandName does not match `/[a-z\-]+` rather than doing replace and stuff.

I would also argue, if we add the key, that we should *not* register the command unless that key is specified. To prefer explicitness over implicitness.

orgkhnargh:
> maybe we should call it atomCommandName instead

I agree.

> we should not register the command unless that key is specified

I agree.

> I think we should throw an error if the commandName does not match /[a-z\-]+/ rather than doing replace and stuff

I don't think so. It's up to Atom to decide if command name is correct or not. In my opinion, we should put a note about how should command names be formatted into `create-provider.md`.

I just went a little deeper and tried to use a function as a command name. Aaaaand... It works just fine.

```js
> commandName = function(){};
> atom.commands.add('atom-workspace', commandName, function(){console.log('apparently, command name should not necessary be a string')})
> atom.commands.dispatch(atom.views.getView(atom.workspace), commandName)
// Prints "apparently, command name should not necessary be a string"
```

Only exception here is that if commandName is an instance of Object, then Atom treats it as a list of commands to register. Otherwise, it just adds it as is.

Also I just realized that we may not add `build:trigger` prefix to the command names, but use `atomCommandName` only. Thus, full responsibility (and control) over the format of command names will be on the provider developer shoulders. What do you think?

noseglid:
I like that idea; register `atomCommandName` as is. In README we can link to above documentation of Atom, but any error is the user's. This also means we make very few assumptions about the API of Atom, which is a good thing. Makes eventual future breaking changes in the Atom API easier to transition to.

orgkhnargh:
See #305

noseglid:
Closed in favor of #305

PR #296, opened 2016-01-23
Check if tool.removeAllListeners esists before call

orgkhnargh:
Some tools may not have `removeAllListeners` method, especially those
which do not extend `EventEmitter`.

Deactivating this package (Settings > Packages > build > Deactivate) leads to the following error: 

```
Error deactivating package 'build' TypeError: tool.removeAllListeners is not a function
    at /path/to/atom-build/lib/build.js:98:12
    ...
```


noseglid:
Awesome. thanks!

PR #305, opened 2016-01-31
Allow to specify a command name for a target

orgkhnargh:
This PR allows to specify a command name and optionally a keymap. Keymap is ignored, unless a command name is specified!

`startNewBuild` now accepts `atomCommandName` as a second argument. Since command names are now set by users (or provider writers), we cannot use them as a source of target names. `startNewBuild` now uses whole command name to find a target to execute. It uses selected target if target with given command name cannot be found.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -3,7 +3,7 @@
 import fs from 'fs';
 import path from 'path';
 import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


noseglid:
Closing in favor of #306 which includes all this and more.

PR #325, opened 2016-02-18
Use cross-spawn when sh is false for better Windows spawning support

johnogle222:
This became a much deeper issue than I had first realized! https://github.com/nodejs/node-v0.x-archive/issues/2318 has a lot of information related to the problem I was having.

Specifically, my home folder on Windows has a space in it. FILE_ACTIVE would provide the full path to the file I was building. But node's escaping isn't ideal when it tries to pass that to the Windows cmd executable.

Based on the numerous references to cross-spawn in the issue above, I ended up implementing this by using cross-spawn but only when sh is false. Using it when sh is true broke a ton of use cases in the spec and I'm sure out in the wild that were implemented like this:
`{
  "cmd": "echo Hello World!",
  "sh": true
}`

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -347,7 +347,7 @@ export default {
 
       this.child = require('child_process').spawn(
         target.sh ? shCmd : exec,
-        target.sh ? [ shCmdArg, [ path.normalize(exec) ].concat(isWin ? args.map(arg => '"'+arg+'"') : args).join(' ') ] : args,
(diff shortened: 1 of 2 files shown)


C-Bouthoorn:
Fixed Travis in https://github.com/johnogle222/atom-build/pull/1

noseglid:
This looks awesome. Thank you so much for taking the time to delve into this! I wanna give this a quick test, which I'll most likely be able to do tomorrow then I'm happy to merge it and make release.

Hooray for better cross-platform support :cake: !

noseglid:
Thank you so much for this! :+1:

PR #327, opened 2016-02-25
Implement pre-/post-build hooks

orgkhnargh:
If target has the "preBuild" attribute and it is a function, it will be called
before the build. If "preBuild" returns a promise, build will be started after
the promise resolution.

"postBuild" hook is called upon child closed. "postBuild" receives a build
status as an argument.

Such feature is necessary for platformio/platformio-atom-ide/issues/45.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -334,7 +334,7 @@ export default {
       }
 
       return Promise.resolve().then(() => {
-        if (target.preBuild && _.isFunction(target.preBuild)) {
+        if (target.preBuild && typeof target.preBuild === 'function') {
           return target.preBuild();
         }
       }).then(() => target);
@@ -417,7 +417,7 @@ export default {
         }
         this.child = null;
 
-        if (target.postBuild && _.isFunction(target.postBuild)) {
+        if (target.postBuild && typeof target.postBuild === 'function') {
           return target.postBuild(success);
         }
       });
--- a/lib/build.js
+++ b/lib/build.js
@@ -333,6 +333,12 @@ export default {
         throw new BuildError('Invalid build file.', 'No executable command specified.');
       }
 
+      return Promise.resolve().then(() => {
+        if (target.preBuild && _.isFunction(target.preBuild)) {
+          return target.preBuild();
+        }
+      }).then(() => target);
+    }).then(target => {
       const env = _.extend({}, process.env, target.env);
       _.forEach(env, (value, key, list) => {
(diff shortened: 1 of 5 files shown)


orgkhnargh:
`it should scan new project roots when they are added` test freezes on dce5d260dd132a226f2d6c23e5b3b476b8192972 as well on my computer. It's not my fault, I guess.

noseglid:
Sorry about the long wait. I have not forgotten about this. I've just had a lot on my plate lately. I will get around to it soon!

lib/build.js
```diff
@@ -333,6 +333,12 @@ export default {
         throw new BuildError('Invalid build file.', 'No executable command specified.');
       }
 
+      return Promise.resolve().then(() => {
+        if (target.preBuild && _.isFunction(target.preBuild)) {
```
noseglid: I want to move away from using lodash, this condition could be reduced to `typeof target.preBuild === 'function'`.

spec/build-hooks-spec.js
```diff
@@ -0,0 +1,110 @@
...
+    waitsForPromise(() => specHelpers.refreshAwaitTargets());
+  });
+
+  afterEach(() => {
+    fs.removeSync(directory);
+  });
+
+  it('should call preBuild', () => {
+    runs(() => {
+      target = getBuildModule().targets[directory].find(t => t.atomCommandName === commandName);
```
noseglid: This uses hidden internal of the build package which increases difficulty of refactoring. I don't think it should be tested this way. Image yourself a user of `build` and what you would expect to input, and what you expect as output.

lib/build.js
orgkhnargh: Sure, will change that.

spec/build-hooks-spec.js
orgkhnargh: Fair point.

In fact, I did initially intend to include small package for this tests, but I thought that it would be better to add a `.json` config as in all other tests to keep the tests consistent.

lib/build.js
orgkhnargh: Done.

spec/build-hooks-spec.js
orgkhnargh: Done.

noseglid:
Very nice! Thank you so much for your contribution. Great work with the specs too! They will really be useful in guaranteeing this functionality.

PR #332, opened 2016-03-04
Added a drag and resize of the build panel

### LINKED ISSUE #312: Build panel size should be draggable
I really would love to see that the build-panel becomes draggable to adjust it's size. I know there is an expanded-mode available but the expanded size which I need often differs, but because it's only configurable in the settings, I often find myself resizing the atom-window and closing/opening the build-panel that is readjusts it's size which is a real pain. 

It also feels just natural that it behaves just like the tree-view or like splitted tabs.

noseglid:
This removes the percentage height. It can now be
set at any height directly.

Also modifies the terminal output so it always
adds show whole rows. I.e. size is rounded to
fit rows.

Fixes #312


noseglid:
Still need to work out Left/Right orientation before releasing.

PR #340, opened 2016-03-14
Use the editor font-family

orgkhnargh:


styles/build.less
```diff
@@ -4,7 +4,7 @@
 .build {
 
   min-width: 250px;
-  font-family: Menlo;
+  font-family: monospace;
```
noseglid: What about having: `'Menlo', 'Consolas', 'DejaVu Sans Mono', monospace` as default rather?

orgkhnargh: I don't think this is required at all, since an actual font-family will be fetched from config, so generally user won't see a font defined here. I think it's fine to set some fancy fonts here just in case something really bad happens :)

noseglid: Defensive programming :)

noseglid:
Thanks for fixing this! Much appreciated!

orgkhnargh:
Glad to help!

PR #351, opened 2016-03-19
[WIP] Variants

robrix:
Many build systems have multiple operations which can be performed on a single target. For example, in addition to building, one might be able to clean, install, test, or profile it. It would be inconvenient to specify these as different _targets_: build providers would often have to duplicate a lot of state to do so, and the user would have to e.g. switch targets to run their tests.

This PR therefore adds support for _variants_, a flexible, opt-in way for build providers to specify alternative operations for a given target.

Each target can contain a `variants` field, which holds an object containing properties which override whatever is set in the base target. It looks something like this:

```js
{
  name: 'All',
  exec: 'stack',
  args: [ 'build' ],
  variants: {
    test: { args: [ 'test' ] }
  }
}
```

A specific variant is selected by the build _source_; if the source starts with `trigger-`, the remainder is the variant. You can see slightly more detail [in my PR against my `build-stack` build provider](https://github.com/robrix/build-stack/pull/2).

I’d further like to change the UI slightly to indicate which variants a given target supports, as well as to indicate the action being taken in the status bar (e.g. “testing target-name” instead of just “target-name”).

For the moment I’ve added a `build:trigger-test` variant to demonstrate this, but I would like to get feedback on the approach before I go too much further. I would like to add support (i.e. well-known commands and keymaps) for `test`, `profile`, and `clean` variants, and possibly `benchmark` as well.

----

- ~~`clean`~~
- ~~`profile`~~
- [ ] Add the current action to the status bar.
- [ ] Indicate available variants in the target switcher.
- [x] Avoid shadowing.
- [ ] Test that:
  - [ ] Triggering an unavailable variant throws.
  - [ ] The triggered variant is copied into the target.


lib/build.js
```diff
@@ -258,6 +263,13 @@ export default {
         env[key] = replace(env[key], target.env);
       });
 
+      if (variant) {
+        const copy = {};
+        Object.assign(copy, target);
+        Object.assign(copy, target.variants[variant]);
+        target = copy;
```
robrix: Since these properties override whatever is set in the target, we make a copy so that building, then testing, then building doesn’t test instead.

C-Bouthoorn: You could also place that note in a comment :wink:

noseglid:
I like the idea. But I would like to propose a variant of the implementation (**pun**!).

I do not think hard-coding variants (such as `test` or `profile`) is a good idea. There will be a million different suggestions of what a "build variant" is from the community and it would spiral out of control.

I would like to see a more generic approach. What if the variants rather inherited all but the specified options and created a new build target all together. For instance

```
{
  "cmd": "buildtool",
  "cwd": "/long/complicated/path",
  "errorMatch": [ "[fancy-regex-stuff]", "[multiple times]" ],
  "variants": {
    "test": { args: [ "test" ] },
    "compile": { args: [ "-c" ] },
    "run": { args: [ "run" ] },
  }
}
```

Implementation wise it would be as easy as looping all the variants and doing `Object.assign({}, root, variant)`.

This could easily be used not only by build providers, but also people writing their custom build files.

What do you think?

lib/build.js
noseglid: Please not that `Object.assign` accepts spread arguments, meaning you can have as many arguments as you want and any keys later will override the earlier ones.

robrix:
@noseglid: While that is admirably simple, it would reduce this to syntactic sugar for defining targets. I can abstract that duplication away in my build provider by writing a function. What my build provider can’t do is give the user a uniform way to build/test/profile/benchmark the active target regardless of what project I’m looking at or what build system it uses.

From my perspective, that’s the important bit of what this PR accomplishes, and likewise it’s the important bit of of what `build` gives me as an end user. So I’d really like to find a way to make this work naturally with `build`.

That said, I definitely agree that everyone will have their own actions which they want to provide. Maybe, in addition to a few common defaults (`test` and `profile`, perhaps, or even just `test`), we could look at which variants the build providers are actually providing and listen for those.

That way, end users don’t need to configure anything at all (unless they’re using custom build files, in which case it’s a simple addition to what they’re doing already). If a user wants a shortcut for an uncommon build variant, they can add one to their keymap like always; and `build` can bless “common-enough” variants by adding default keymaps, if so desired.

In fact, we wouldn’t have to hard-code any variants at all then; I’d be more than happy to add shortcuts to my own keymap, and suggest that `build-stack`’s users do the same.

robrix:
@noseglid: I’ve updated the PR to implement the adjustment to the proposal made above :point_up: It’s working quite well, and you can even find the variants in the command palette.

noseglid:
I'd want to review and test this, but don't wanna do so until the daily builds are working. Are you looking into that?

robrix:
> I'd want to review and test this, but don't wanna do so until the daily builds are working. Are you looking into that?

My bad, I was waiting for response to my revised proposal :grin: I’ll resolve the CI failures.

robrix:
@noseglid: This should pass the tests now.

noseglid:
Cool! I'll have a look at it shortly. Currently on a trip!

robrix:
Enjoy your trip! :blush:

lib/build.js
```diff
@@ -182,6 +183,27 @@ export default {
...
+          const commands = Object.keys(newTarget.variants).reduce((table, variant) => {
+            const triggerName = 'trigger-' + variant;
+            table['build:' + triggerName] = () => this.build(triggerName);
+            return table;
+          }, {});
+          this.variantsSubscriptions = atom.commands.add('atom-workspace', commands);
+        }
+      });
+    }
+  },
```
noseglid: I don't quite get this. Why would `setActiveTargetName` register atom commands?

noseglid:
I visioned variants to be registered as normal commands. That way, `atomCommandName` can be used to optionally register a command name. It would have the same effect as what you're doing now, with the added benefit that the variants will be visible in the available targets list view.

I know above is diverging a bit from what you originally intended with this PR (a uniform way to run various tasks - such as test and profile - associated with a target). I just don't think that should be in the scope of this plugin, which is why I am trying to find a middle way with something I do think is within the scope of this plugin **and** can help you solve your challenge :smile: 

Any reason why you're not allowing `variants` to be used through the custom build file implementation?

There must also be some documentation and specs with this.

robrix:
> I visioned variants to be registered as normal commands. That way, `atomCommandName` can be used to optionally register a command name. It would have the same effect as what you're doing now, with the added benefit that the variants will be visible in the available targets list view.

As I said before, that’s in direct conflict with what I’m trying to achieve. I very much _do not_ want to have multiple targets for different actions, one target per action—if I did, I could do that right now without variants, by just making more targets in my build provider. Rather, I want multiple actions per defined target, in much the same way that e.g. an Xcode target can be built, cleaned, run, archived, profiled, tested, etc.

> I know above is diverging a bit from what you originally intended with this PR (a uniform way to run various tasks - such as test and profile - associated with a target). I just don't think that should be in the scope of this plugin, which is why I am trying to find a middle way with something I do think is within the scope of this plugin **and** can help you solve your challenge :smile:

If you don’t consider multiple actions per target to be in scope, that’s obviously your decision, and I’ll respect that. But in that case, I don’t think this PR offers anything new or valuable—build providers can already make new targets conveniently by the use of a “template” function, without requiring `build` to add any special support. So I’m going to close this PR.

Again, I respect your decision to limit `build`s scope to a single action per target. But as an end user, I’m quite disappointed. This complicates my workflow quite a bit, and limits the value I get from `build` at all.

Nevertheless, thank you for reviewing!

noseglid:
I am very sorry about all the effort you put down that didn't lead to anything.

I just want to be clear - I don't think it's a bad idea. Just not in scope.

Thanks again for your time.

robrix:
@noseglid: Thanks for circling back, I appreciate it. These things happen :blush:

PR #353, opened 2016-03-21
Added watchr to fix noseglid/atom-build#277

C-Bouthoorn:
Added watchr to fix https://github.com/noseglid/atom-build/issues/277

### DIFF
--- a/lib/atom-build.js
+++ b/lib/atom-build.js
@@ -36,21 +36,25 @@ export default class CustomFile extends EventEmitter {
   constructor(cwd) {
     super();
     this.cwd = cwd;
-    this.filewatcher = undefined;
+    this.filewatcher = undefined;  // Mac and Linux
+    this.fileWatchers = [];  // Windows
   }
 
   destructor() {
+    // Mac and Linux
     if (this.filewatcher && this.filewatcher.close) {
       this.filewatcher.close();
     } else {
-      const details = this.filewatcher ? 'Close function is not defined' : 'Filewatcher is not initalized';
-
       atom.notifications.addError('Failed to close filewatcher', {
-        detail: details,
+        detail: this.filewatcher ? 'Close function is not defined' : 'Filewatcher is not initalized',
         dismissable: true
       });
     }
-    // this.fileWatchers.forEach(fw => fw.close());
+
+    // Windows
+    if (this.fileWatchers) {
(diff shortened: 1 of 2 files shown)


C-Bouthoorn:
*Sidenote: Funny how Travis passes these checks, while it clearly fails?*

C-Bouthoorn:
Now it works for Linux and Mac, but it doesn't for Windows. This looks like a problem with watchr...

@noseglid Should I try fixing watchr, add an os-check before using watchr, or search for another library?

noseglid:
There's definitely something weird going on with windows.
Maybe it'd be better to just do the OS check and use `watchFile` on windows and try to work out the caching issue?

C-Bouthoorn:
Watchr fixes the caching issue, so that is not the matter anymore. I'll add an OS check

noseglid:
One could question the use of `watchr` if we do OS corrections ourselves. After all it's an additional dependency which increases size and load times.

Maybe we could scavange the cache clearing solution from `watchr` and copy that?

C-Bouthoorn:
To me, `watchr` looks a bit complicated. If someone else would like to search for the responsible code, please. In the meantime, I'll try to get this to work. Maybe later someone will be able to speed up the loading progress, but for now, I think fixing *this* bug is more important.

noseglid:
I'm sorry for taking so long to respond to this. I don't feel comfortable merging this with all these unknowns:

  * We don't know why `watchr` doesn't work on windows.
  * We don't know why `watchr` doesn't crash on linux.

I really feel like we need to sort this out so we can hopefully avoid the `watchr` dependency, or use only `watchr` on all platforms.

I'll help you look into this as soon as I have some time if you don't find anything. I'm very eager to find a fix for this issue - so it's not like I don't want to fix this.

C-Bouthoorn:
Hey! Sorry I was inactive this week. I will see which functions `watchr` uses here, and try to mimic those in build. Still don't have much time today, but I'll keep it in my mind.

C-Bouthoorn:
Reading all other commits. Doesn't https://github.com/noseglid/atom-build/pull/372 fix this exact issue?

noseglid:
Exactly. That's my alternative solution so we don't need `watchr`. Question is, does it solve the crash for you?

C-Bouthoorn:
As stated on that commit, it indeed does. Closing this issue

PR #356, opened 2016-03-24
Don't normalize cmd

jhasse:
Fix execution of commands on Windows which contain `/`.

Try the following `.atom-build.yml` on Windows:

```
cmd: explorer https://github.com/noseglid/atom-build
```

Expected result: Opens the correct URL in the browser.
Actual result: Tries to open `https:\github.com\noseglid\atom-build`.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -250,7 +250,7 @@ export default {
       if (target.sh) {
         this.child = require('child_process').spawn(
           shCmd,
-          [ shCmdArg, [ require('path').normalize(exec) ].concat(args).join(' ')],
+          [ shCmdArg, [ exec ].concat(args).join(' ')],
           { cwd: cwd, env: env }
         );
       } else {


noseglid:
I've tracked the place it was introduced to: 2a6f66eeea340c9201cfa3a90d4fea15cd575384

I'm inclined to agree with you that it shouldn't be normalized, altough I am uncertain which effects it might have. @dreamair do you have any insight?

dreamair:
How about putting the URL into the arguments (args)? The command (cmd) should be a file, shouldn't it?

noseglid:
That'd be a work around. But why is it normalized in the first place?

dreamair:
A work around? What do you have the 'args' for, if you want to interpret the 'cmd' as a complete command line with all the arguments?
If the command file includes a path, eg. node_modules/.bin/somecmd, it should correctly run node_modules\.bin\somecmd on Windows.
A complete cross-platform solution would probably also normalize the arguments to correctly handle file paths. URL arguments should then be put between double-quotes, I guess.
Personally I only deal with non-path 'cmd's so far and I don't mind you removing the normalization. I still think the normalization at least on the 'cmd' and using 'args' for the arguments is a better way to go.

jhasse:
@dreamair If `sh: true` I want to be able to do stuff like `build_command && second_command some_arg && third_command another_arg`.

dreamair:
That looks to me more like a build script and I wouldn't put that in a build configuration for an IDE. I would keep this atom-build simple and well, platform independent.
If not so I would suggest a non-cross-platform 'cmdLn' spec replacing the 'cmd' and 'args' for such cases.
I use npm run, gulp and/or simple nodejs apps for build scripts.

jhasse:
That would be a major inconvenience for me. Can you give me an example of a command where you want to have the path normalized? Why not use `cwd` for that?

dreamair:
Cwd is the working directory, not the path to the command. And see a previous post for an example.

jhasse:
Couldn't find the example in your previous post.

Aren't `npm run`, `gulp` global commands which wouldn't need to be normalized?

dreamair:
Yes, if commands are in the global system path, they can be called directly. But my "generic" example in the previous post uses a relative path ('node_modules/.bin/someCmd'). When you want to have a per-project only dependency installed, 'npm install someCmd' without the '-g', the command is only available in the node_modules/.bin/ folder. Since you still want to have your working directory at the root of the project, you need to specify the path for the command. And in order for paths to be cross-platform they need to be normalized.

dreamair:
I think, the base question here is, where the future focus for atom-build should be: 1) a build plugin for the atom editor with a simple, straight forward, cross-platform configuration or 2) a build script engine, replacing system dependent shell scripts, Gulp, Grunt,... and integrating with the editor as a side effect ;) ...sorry, maybe a little too black&white ;)

my vote is for 1) and I would even like the root 'cmd', 'cwd' and 'args' be a default for all targets. In our configs a target mostly defines only an additional argument. I think, tasks.json in vscode offers something similar.

jhasse:
What do you think about only normalizing until the first space?

dreamair:
Well that would be a work around, an ugly one ;) Good it's not up to me to decide!

PS: @jhasse For you another work around might work as well: try putting your cmd1 '&& cmd2 ...' in the first 'args' string and only the first cmd1 in the 'cmd'. Even more ugly, but since the args are not normalized yet it might work ;)

noseglid:
Good discussion. I'm thinking it might actually make sense to only normalize the first part of cmd. However it might be tricky to determine what the first part is since spaces can be part of a path. 

Maybe we should only normalize cmd if sh is false - in that case cmd is expected to be the executable only. 

I'm away during Easter but I'll try to be as responsive as possible

noseglid:
Also regarding args, cmd and sh - this is just really mirroring the node spawn api

dreamair:
Good morning ;) Sorry, I should have explained my ugliness reaction, but I just don't like public interfaces behaving "funny". So 'tools/cmd1 && cmd2' runs also on Windows, but 'cmd1 && tools/cmd2' doesn't? Please make sure to document such "features"!

dreamair:
The Node spawn API is placed in an environment where you have control over process.platform === 'win32'... and even with that different context they still provide a separate exec API where you can pass a command line and no args. 
A choice between 'cmdLn: "cmd1 && cmd2 > out.txt"' (or 'exec'?) for a command line or 'cmd: "explorer"' and 'args: ["http://..."]' would be clearer to me.

jhasse:
You're right, only normalize until the first space is a bad idea.

What about an option `normalizePath: true/false`?

The way it works now is confusing, as it shows the un-normalized cmd in the build panel.

noseglid:
valid points @dreamair. 

I'm wondering if it really should be cross-platform. If someone creates a build file it's very likely that it will be for a single platform ( in which case backslashes can be used instead on windows). For build providers they have access to `process.platform`. 

The only time that wouldn't work is if someone is working on a cross platform application and commits the atom-build files to work across platforms. 

So what I am suggesting is to just never normalize and let the user do it. Possibly normalize if `sh` is true because it's guaranteed that `cmd` is only the command in this case. 

Alternatively what @jhasse is suggesting, which would solve the above use case too, at the expense of increased configuration complexity

dreamair:
In my experience many teams work on Windows and Mac/Linux. And yes they like to commit the whole setup. Now with nodejs, atom,... running on both quite well.
While normalizePath: true/false would work for this case, but I find the cmdLn cleaner. If someone needs to write scripts into cmd, the args are not needed, are they?
Hmm, I still don't get it, what's so difficult to write a separate script and call it from the atombuild? In there you have the whole power, work with syntax highlighting,... And you can even put it in a tools folder, one for Unix and a *.cmd one for Windows ;)
As I mentioned earlier, do you want to offer yet another build system or an atom plugin with a simple config?

jhasse:
> Hmm, I still don't get it, what's so difficult to write a separate script and call it from the atombuild? In there you have the whole power, work with syntax highlighting,... And you can even put it in a tools folder, one for Unix and a *.cmd one for Windows ;)
> As I mentioned earlier, do you want to offer yet another build system or an atom plugin with a simple config?

The same could be said about your use-case: Just put a script which handles the path normalization next to your `.atom-build.yml` and your good to go. Do you want to offer yet another cross-platform tool which changes characters under your feed or an atom plugin with a simple config (which just executes what you tell it to)?

dreamair:
Good reply! ;) 
hmm, but the atom editor is cross-platform. And wouldn't you expect a plugin for it to be usable in the same way?

noseglid:
I think there must be a limit to what cross platform functionality build provides. For instance `echo` exists in both windows and Linux/UNIX but behave differently. Build won't accommodate for that. 

If `sh` is true all of `cmd` and `args` is passed to the appropriate shell. Any platform considerations must be taken by the user here. If `sh` is false then cmd is guaranteed to be path to command only, as this is how exec/spawn is defined. We can help the user here by normalizing. 

we should document this in the readme

@jhasse will you make a conditional check for sh befor deciding if to normalize?

dreamair:
Another advantage of keeping build scripts separate from tool configurations is the ability to run them independently from the tool. You cannot run a script for example directly from the command line if it's in the atombuild. Or one of your team uses sublime instead of atom.
A tool configuration like this should only map actions to actual scripts of some kind in the simplest, most manageable way. Everything more is reinventing the wheel and would ask for more power like platform conditionals, parallel execution, task dependencies,... So I'd rather have my scripts in a build/ or tools/ folder organized and not having to bother who is working on Windows, than writing whole scripts in a limited context like the atombuild configuration and/or be limited to place all my scripts in the project root folder.

dreamair:
? is 'sh' not the flag for opening a shell instead of directly running a process?

noseglid:
Exactly. If sh is true then there may be arguments in cmd ( such as a url to `open` as described by

noseglid:
Oops wrong button.  

... As described by @jhasse )

dreamair:
I think opening a shell or having arguments or not are different things. For example in the initial example 'explorer' would not require a shell to run, but some scripts do.
If you really don't care about cross-platform, I would vote for removing the normalization at all. At least then it stays simple.

noseglid:
I don't see any harm in normalizing for cmd if sh is false. It can help someone be crossplatform if they move the executable only to cmd and args in rest

dreamair:
Harm one: funny behaviour again, sh: true/false has nothing to do with cross platform. I don't think anybody will naturally expect one being normalized while the other is not. Harm two: shell scripts, at least on Windows need a shell to run. So you wouldn't be able to run build/someScript on Windows this way.

noseglid:
Regarding 1: It's not about cross platform. It's about helping out where we can. How does this harm?

Regarding 2: Shell scripts (e.g. `.bat` or similar) do need a shell to run. So in this case you'd need to set `sh: true`. For other commands, you don't need a shell, which is where normalization can help.

You could also set `sh: false`, `cmd`: `cmd.exe` and args to the shell script (Which is essentially what `sh: true` does.

I still don't see the harms

dreamair:
1) You are using a config flag to produce side effects. What does shell or not shell have to do with normalized paths? You don't help anyone by making things confusing.
2) I cannot remember a setup having a local command not being a script. Processes on the other hand are normally globally accessible. So for most cases the normalization would help (having scripts in subfolders) your solution would not help.

noseglid:
So you're in favor of merging this PR as is (E.g. removing all path normalization) ?

dreamair:
eastern family duties, phu.. ;)

Sorry for repeating myself again: I'm in favor of separating build scripts and tool configuration. This means here keeping the code as it is. And maybe add some FAQs, something like:

"cmd: explorer https://github.com/noseglid/atom-build" opens the wrong URL! What can I do?
Please use the 'cmd' for the command only and 'args' for the arguments! 
"cmd": "explorer", "args": ["https://github.com/noseglid/atom-build"]
For cross-platform support any path in the 'cmd' specification will be normalized.

How do I specify build scripts in the atom-build config? I want to write "cmd": "build_command && second_command some_arg && third_command another_arg".
You shouldn't. Please use separate shell scripts or build tools to specify your build scripts!
"cmd": "build/myScript" and build/myScript containing "build_command && second_command some_arg && third_command another_arg". Your script might be different for Windows. Then simply add that to the build/myScript.cmd file.

jhasse:
This behavior will result in rather unexpected results. The other way - no normalization - makes more sense to me. If someone executes `cmd /C build/script/some_cmd` by himself on Windows he should get the exact some result as within Atom! And Windows people know about `/` not working, so no need to document things.

> Cwd is the working directory, not the path to the command.

So? You can execute `build\script\windows.cmd` via:

```
cmd: windows
cwd: build/script
```

And I think normalizing `cwd` makes a LOT more sense, since it is a directory.

jhasse:
Also regarding the cross-platform concern: I just tested: `cmd: xdg-open https://github.com/noseglid/atom-build` works on Linux. So if someone passes a URL to his script via `cmd` this will only NOT work on Windows.

dreamair:
"This behavior will result in rather unexpected results." What do you mean by "This" and what results are expected?
As for your cmd example: Atom-build should configure build targets not duplicate the Windows/Linux terminal.
For the cwd: this way all paths in windows.cmd must first compensate for it's location (../../src,...). Some tools require a specific cwd to work... just discovered the Typings (tsd replacement for TypeScript is such a tool ;( . Again please use one parameter for one clear purpose only. The working directory and the directory the script is placed in are two different things. 
But I agree with normalizing cwd!
I would have guessed URLs in cmds also not to work on Linux. Try cmd: xdg-open, args:[https://github.com/noseglid/atom-build].

jhasse:
> What do you mean by "This" and what results are expected?

With "this" I mean the normalization of cmd. I expect `cmd: explorer https://github.com/noseglid/atom-build` to work and `cmd: build/script.cmd` to fail (just as it does in cmd.exe).

> For the cwd: this way all paths in windows.cmd must first compensate for it's location (../../src,...). Some tools require a specific cwd to work... just discovered the Typings (tsd replacement for TypeScript is such a tool ;( .

Use `cd ..\..` in the first line of the `.cmd` file to fix this.

> But I agree with normalizing cwd!

That's already implemented and wouldn't be changed by this PR ;)

> I would have guessed URLs in cmds also not to work on Linux. Try cmd: xdg-open, args:[https://github.com/noseglid/atom-build].

Works too :)

noseglid:
I am unwilling to remove the functionality to allow arguments in `cmd` since it may break a lot of people's configurations. But I think we can agree that the solution proposed by @jhasse is acceptable as is.

dreamair:
...and using the build/script/windows from the project root would break. Only because you want to include a build script directly in a config file!

noseglid:
Yes. I've gotten your point, @dreamair and I don't agree with it. While it may have been beneficial to implement it like that from the beginning, that's not where we are now.

dreamair:
Yes, I can see that point. And luckily you don't need my agreement, do you ;)

jhasse:
In order not to break people relying on the normalization on Windows, we could at least change it so that the normalized path is displayed in the build view?

@dreamair is right that passing a URL on Windows can be achieved via `args` and the more I think about it, the more I agree that this PR shouldn't be merged.

noseglid:
You don't think it's weird that it normalizes the arguments?

jhasse:
It's weird, yes. But breaking the build files of people via an update also isn't good :/

noseglid:
That is indeed never good. There wouldn't be an easy way to achieve the same functionality if this PR is merged either.

noseglid:
I will merge this as it is. We can always revert this if it causes too much trouble, but I suspect most won't notice a thing.

PR #360, opened 2016-03-25
Live error match

farfromrefug:
In my dev env (Titanium Mobile), we run javascript interpreted code on the iOS/Android simulators.
Which means that error can occur in real time.
Also the build never really end until we stop the simulator.
So it means that error wont get catched in the system used now.

So i added a target option ```liveErrorMatch``` which will allow live error matching.
I created the pull request against the ```linter-integration``` branch to test it all.
I can report that it works very well!

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -221,6 +221,9 @@ export default {
       range: [[match.line - 1, (match.col || 1) - 1], [match.line - 1, (match.col || 1) - 1]]
     }));
     this.linter.setMessages(matches);
+    if (matches.length > 0 && atom.config.get('build.scrollOnError')) {
+      this.errorMatcher.matchLast();
+    }
   },
 
   startNewBuild(source, atomCommandName) {
--- a/lib/build.js
+++ b/lib/build.js
@@ -286,11 +286,10 @@ export default {
       let stdout = '';
       let stderr = '';
       let liveErroMatchQueue = '';
-      var liveErroMatchHandle = target.liveErrorMatch ? function(d) {
+      let liveErroMatchHandle = target.liveErrorMatch ? function (d) {
         d = liveErroMatchQueue + d;
         liveErroMatchQueue = '';
-        var index = d.lastIndexOf('\n');
-        console.log('liveErroMatchHandle', liveErroMatchQueue, d.length, index);
+        let index = d.lastIndexOf('\n');
         if (index >= 0 && index === d.length - 1) {
           that.errorMatcher.parseAdd(d);
         } else {
@@ -298,14 +297,14 @@ export default {
           that.errorMatcher.parseAdd(d.substring(0, index));
         }
         that.updateErrorLinter();
-      } : function() {};
+      } : function () {};
       this.child.stdout.setEncoding('utf8');
       this.child.stderr.setEncoding('utf8');
-      this.child.stdout.on('data', function(d) {
+      this.child.stdout.on('data', function (d) {
         stdout += d;
         liveErroMatchHandle(d);
       });
-      this.child.stderr.on('data', function(d) {
+      this.child.stderr.on('data', function (d) {
         stderr += d;
         liveErroMatchHandle(d);
       });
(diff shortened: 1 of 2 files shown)


C-Bouthoorn:
I want to fix these errors on your branch, but I can't access it. Could you send me a link so I can fix them?

C-Bouthoorn:
Ehm, it looks like the patch-1 isn't merged?

noseglid:
This is an interesting idea. Couple of questions:

* How does it work with `Automatically scroll on build error` option set?
* The buffers received from `stdout` and `stderr` has no size guarantee. The callback can be one byte each. What happens if an error match is split between two buffers?

farfromrefug:
I now handle both.
about ```Automatically scroll on build error``` it will automatically scroll to ```last``` error when a new error is found.

noseglid:
You need to re-work this PR a bit.

  * Fix the lint errors.
  * Look a the structure of the code. Don't put error matcher functionality in `build.js`. Keep the interfaces clean.
  * Add tests, add documentation.

farfromrefug:
Sure i will look at this:
* what error matcher functionnality did i put in ```build.js```? I don't see it.
* i am new to atom tests and stuff. It needs to be in spec, right?

lib/build.js
```diff
@@ -266,10 +285,29 @@ export default {
 
       let stdout = '';
       let stderr = '';
+      let liveErroMatchQueue = '';
+      let liveErroMatchHandle = target.liveErrorMatch ? function (d) {
```
noseglid: Why is this conditioned by the target? Any reason to not alway do live error matching?

lib/build.js
```diff
@@ -266,10 +285,29 @@ export default {
...
+        d = liveErroMatchQueue + d;
+        liveErroMatchQueue = '';
+        let index = d.lastIndexOf('\n');
+        if (index >= 0 && index === d.length - 1) {
+          that.errorMatcher.parseAdd(d);
+        } else {
+          liveErroMatchQueue = d.substring(index + 1);
+          that.errorMatcher.parseAdd(d.substring(0, index));
+        }
```
noseglid: Now you're adding lines, but there are many reasons why this doesn't work:

  * An error match may span many lines
  * Newlines are different on different OSes and also different tools within OSes (e.g. running cygwin on Windows may have a tool outputting `\n` as new lines even though `\r\n` is the line separator on windows). Build can't make any assumptions on the line endings.

noseglid:
I'm actually not sure the desired functionality can be achieved with current structure.

I think the error match need to continually evaluate the full output, which will not work very well if - as you describe - there is a long running target which never finishes.

lib/build.js
farfromrefug: Well yes, 
If i have liveErrorMatch i don't want to parse again when finished. This actually does not make sense.
For example in my case i run my app through iOS simulator. When ending the session it means i either stopped to build by choice(esc) or closed the simulator. I don't think at that point i need to parse the stdout/stderr again.

Which means with this target (Titanium Mobile on iOS simulator) i want liveErrorMatch and not "end" error Match.
But i still don't want it when building with make or cmake for example.
So i guess it is a target parameter

lib/build.js
farfromrefug: But there is always a ```\n```, right?

noseglid: Consider this error message from `eslint`:
```
/Users/noseglid/devel/atom-build/lib/atom-build.js
  62:29  error  Missing semicolon  semi
```

That's multiple lines.

lib/build.js
noseglid: So why do end matching at all? If we do live error matching we might as well skip the final step (since the last data buffer dropping in will cause the error matching to run).

lib/build.js
farfromrefug: That's an actual error output?
Anyway you are right we can't handle all cases. I actually think that impossible.
But you will still handle most of them without breaking anything!
And it won't have weird behaviors.
Refuse it if you want but i am sure i am not the only one who could use that (all interpreted compilation languages could use that)

PR #361, opened 2016-03-25
don't rely on editor path

farfromrefug:
I faced that issue where i was editing a file outside of the project and the build was not working.
I think the build should be related to the opened project and not the editing file.

I am not sure of the right way to fix that but that dirty fix make it works in all cases

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -25,9 +25,8 @@ const activePath = () => {
       return false;
     }
 
-    return atom.project.getPaths()[0];
   }
-
+  return atom.project.getPaths()[0];
   /* otherwise, build the one in the root of the active editor */
   return atom.project.getPaths().sort((a, b) => (b.length - a.length)).find(p => {
     try {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -24,11 +24,10 @@ const activePath = () => {
     if (0 === atom.project.getPaths().length) {
       return false;
     }
-
+    return atom.project.getPaths()[0];
   }
-  return atom.project.getPaths()[0];
   /* otherwise, build the one in the root of the active editor */
-  return atom.project.getPaths().sort((a, b) => (b.length - a.length)).find(p => {
+  var result = atom.project.getPaths().sort((a, b) => (b.length - a.length)).find(p => {
     try {
       const realpath = fs.realpathSync(p);
       return textEditor.getPath().substr(0, realpath.length) === realpath;
@@ -37,6 +36,9 @@ const activePath = () => {
       return false;
     }
   });
+  //if we don't find a project path containing the edited file then
+  //return the first project path as default
+  return result || atom.project.getPaths()[0];
 };
 
 const getDefaultSettings = (cwd, setting) => {


orgkhnargh:
This only helps when you have exactly one project open. With this change you will not be able to run builds against any other projects in a tree view, except for the first one, which will make `build` useless when you edit more than one project in a single Atom window.

farfromrefug:
Well i don't think atom-build support multiple projects, does it?
But i agree, and i mentionned, this is not the "right" solution. But at least it works better in most cases i think.

orgkhnargh:
`atom-build` does support multiple projects. The function you've modified is responsible for determining which project is active at the moment. Basically, it searches for what project does current open file belong to. You just made it return the first open project, no matter what file is open, which is definitely not the right solution. Your solution only works better in one case - when you have only one project per window. In this case you are able to edit files outside of the open project, but the project will still be considered active, thus `atom-build` functionality will remain available when the non-project file is active.

What would really help here is the way to choose the active project explicitly. For example, via a select panel, like one for choosing a target. But I don't think it is necessary. I think, it's better to just add a section to the README.md with description of how the active project is selected, so nobody will get confused.

farfromrefug:
Oh i see so the problem might be only when the editor file is NOT within the projects path. In this case it should default to returning the first project path may be?
At least that would be a better fix for now.

As for selecting the default build project i agree it should default to the first one and allow you to change it

orgkhnargh:
> i agree it should default to the first one and allow you to change it

Actually, I don't think this is necessary. Activating a file from the project I want to build right before building it is what people normally do, don't you think so? First you open a file, then do some changes, then trigger a build and wait for a result.

noseglid:
The build plugin is project oriented. I don't want to change that at this point because I think it opens too many unknowns. Just defaulting to the first project will lead to some pretty weird behavior and not make much sense to a lot of users. If we want to move `build` away from its project orientation we need a bigger design overhaul I think.

farfromrefug:
Well to be honest you say it could lead to weird behavior. What it is doing right is really weird.
It is project oriented you are right !
But when you open an "outside" file like a readme or anything it should still build the project.
Or right now it wont! It won't be anything if the current edited file is not part of the project.

noseglid:
You're right. It wont be anything. The build files `.atom-build.json`, `.atom-build.yml` etc. are put in the root of a project.

Now we're editing a file which is in a completely different location on disk. Which project should atom-build look to for configurations? There's no relation between the open file and a build configuration.

Arbitrarily looking at the "first open project" just increases confusion in my opinion. It is better to keep atom-build strictly project oriented.

farfromrefug:
I get your point. But just so you see my point it took me 3 hours to understand why the build was not triggering... Until i got deep into your code and see that my editing file was not part of the project.
I come from Sublime Text and projects are handled a little differently (one project per window) which makes it easy(er).
Anyway this is your project but you should at least make it an option to "auto selected project". I am sure i am not the only user that would get totally frustrated by that.

PR #365, opened 2016-03-27
Support for `.atom-build.js`

### LINKED ISSUE #362: [FR] add some kind of postBuild event for .atom-build files
I am not how this would be implemented

Here s my issue, i would like some javascript code to be run on post build.
Especially this:
```javascript
            if (success) {
                atom.notifications.addSuccess('Build successfull!');

            } else {
                atom.notifications.addError('Failed to build.');
            }
```

This is doable when creating a custom build provider but not possible when using .atom-build file.
I wonder if this could get done in any way.

May be by adding support for ".atom-build.js" which we would evaluate in ```getConfig```?

noseglid:
Support for defining `.atom-build.js`.

This format support functions so that `preBuild` and `postBuild` can be defined.
The configuration must be `module.exports`'d.

Need

  * [x] specs
  * [x] documentation

Fixes #362

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Supported formats and the name of the configuration file is
   * JSON: `.atom-build.json`
   * CSON: `.atom-build.cson`
   * YAML: `.atom-build.yml`
+  * JS: `.atom-build.js`
 
 Pick your favorite format, save that file in your project root, and specify exactly
 how your project is built (example in `json`)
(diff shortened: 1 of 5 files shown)


noseglid:
Need

[ ] specs
[ ] documentation

noseglid:
Need

  * [ ] specs
  * [ ] documentation

PR #369, opened 2016-03-30
support relative paths in linter + spans that span multiple characters/lines

### LINKED ISSUE #359: errorMatch: option to capture character range
Alright, this might sound odd. I've come across a compiler that does not report the error in the common `<line>:<col>` scheme, but the character range in which the error occurred.

**Example**:

Here is an AppleScript that will break on build

```applescript
tell application "Finder"
  if condition --then omitted
    -- statements
  end if
end tell
```

Building would produce the following error

```
/Users/me/Desktop/broken.applescript:31:40: execution error: The variable condition is not define
d. (-2753)
```

The number `48` marks the beginning of the range where the error occurs, `57` marks the end.

I'm not sure how many compilers follow this, but it would be nice to have a pattern that allows us to capture this, e.g. `?<from>` and `?<to>`.

oli-obk:
previously the lints did not show up in files if the paths were relative. Alternatively this could be changed in atom-linter directly, but I don't think that's feasible as atom-linter doesn't know about projects, but atom-build explicitly does.

fixes #359

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -300,10 +300,11 @@ export default {
           success = success && !this.errorMatcher.hasMatch();
         }
 
+        const path = require('path');
         this.linter && this.linter.setMessages(this.errorMatcher.getMatches().map(match => ({
           type: 'Error',
           text: match.message || 'Error from build',
-          filePath: match.file,
+          filePath: path.isAbsolute(match.file) ? match.file : path.join(cwd, match.file),
           range: [
             [ (match.line || 1) - 1, (match.col || 1) - 1 ],
             [ (match.line || 1) - 1, (match.col || 1) - 1]
--- a/lib/build.js
+++ b/lib/build.js
@@ -307,7 +307,7 @@ export default {
           filePath: path.isAbsolute(match.file) ? match.file : path.join(cwd, match.file),
           range: [
             [ (match.line || 1) - 1, (match.col || 1) - 1 ],
-            [ (match.line || 1) - 1, (match.col || 1) - 1]
+            [ (match.line_end || match.line || 1) - 1, (match.col_end || match.col || 1) - 1 ]
           ]
         })));
 
--- a/README.md
+++ b/README.md
@@ -157,8 +157,10 @@ matched by the regular expression `RE`.
 
 The following named groups can be matched from the output:
   * `file` - **[required]** the file to open. May be relative `cwd` or absolute. `(?<file> RE)`.
-  * `line` - *[optional]* the line the error resides on. `(?<line> RE)`.
-  * `col` - *[optional]* the column the error resides on. `(?<col> RE)`.
+  * `line` - *[optional]* the line the error starts on. `(?<line> RE)`.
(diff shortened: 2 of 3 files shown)


lib/build.js
```diff
@@ -294,13 +294,14 @@ export default {
           success = success && !this.errorMatcher.hasMatch();
         }
 
+        const path = require('path');
         this.linter && this.linter.setMessages(this.errorMatcher.getMatches().map(match => ({
           type: 'Error',
           text: match.message || 'Error from build',
-          filePath: match.file,
+          filePath: path.isAbsolute(match.file) ? match.file : path.join(atom.project.getPaths()[0], match.file),
```
noseglid: The error match file/configuration is not necessarily in the first project. The path you're actually looking for here is in the variable `p` (determined at the top of this function).

lib/build.js
```diff
@@ -294,13 +294,14 @@ export default {
...
           range: [
             [ (match.line || 1) - 1, (match.col || 1) - 1 ],
-            [ (match.line || 1) - 1, (match.col || 1) - 1]
+            [ (match.line_end || match.line || 1) - 1, (match.col_end || match.col || 1) - 1 ]
```
noseglid: Nice, could you also add this to the README.

Now that I am seeing this, the whole linter integration thingie should probably have been in `error-matcher.js` insteaed. Not that that has anything to do with your PR :)

oli-obk:
addressed comments by fixing the appropriate commits

noseglid:
What is the problem with relative paths? It works well for me:
error-match

Possibly it should be relative `cwd` instead? As that is the path that the error is most likely to be relative to.

oli-obk:
the problem is that you don't get a hit on `File` only on `Project`. You can seen this at the bottom of the screen. With the changes implemented in this PR, you also get a squiggly line at the line+col location and a popup message showing the error message (if any). With the changes I made to cargo, you get a squiggly line below the entire error location (this can be applied to gcc errors and many others, too).

noseglid:
Still something weird with the test: https://travis-ci.org/noseglid/atom-build/jobs/119506163

Also, I think it makes more sense to take it relative to the cwd of the executed command (rather than active project path).

noseglid:
This will solve #359

oli-obk:
> Also, I think it makes more sense to take it relative to the cwd of the executed command (rather than active project path).

oh. that makes sense

noseglid:
`cwd` is available in that scope, unless I am mistaken. So should be as easy as changing that

And the remaining failing spec probably has the same fix as the one you already corrected. (You can run specs locally in your `Window: Run Package Specs`). They require no setup.

oli-obk:
changed to use `cwd` and fixed the test

noseglid:
The slashes `/` in the specs should be backslashes `\` on windows. Use `path.join` to get the correct delimiter.

oli-obk:
travis and appveyor are happy now

noseglid:
Very nice. Thanks for this!

PR #371, opened 2016-03-30
.atom-build targets properties default to main target

farfromrefug:
When using targets in a .atom-build file, it seems logical for the targets to "inherit" the main target.
This also makes thing a lot easier by preventing duplicate errorMatcher, postBuild, env,...

### DIFF
--- a/lib/atom-build.js
+++ b/lib/atom-build.js
@@ -1,6 +1,7 @@
 'use babel';
 
 import EventEmitter from 'events';
+import _ from 'lodash';
 
 function getConfig(file) {
   const fs = require('fs');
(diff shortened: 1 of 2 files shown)


noseglid:
I am sorry but I do not want this functionality. Here are my reasons:

  * Introducing this now could potentially break a lot of build since it would start inheriting where it previously did not (e.g. `cwd` might propagate to a target which does not have it defined).
  * This is nothing but syntactic sugar, but it increases the complexity of the project, and unnecessarily so.
  * For this to actually be mergable, additional constructs - such as `unset` - would be required.

(Also, I [recently spent quite some effort](https://github.com/noseglid/atom-build/commit/042f7a55f6148a0b503f8d0904314e371702d7e8) to get `lodash` out of the project.)

Thank you for your effort though.

PR #373, opened 2016-03-30
Added /opt/local/bin as a possible path extension.

werner291:
Hi!

On my OSX machine, the command-line version of cmake was installed in "/opt/local/bin", which your package couldn't find.

This change fixes that.


noseglid:
Thank you for the PR, but you should modify your PATH setting outside of the package. The reason `/usr/local/bin` is in there is legacy only and I will remove it as soon as Atom 1.7 hits. Atom will read your shell settings at that point.

PR #383, opened 2016-04-07
Issue with 'Panel Orientation' Left or Right

abartelsmeier:
Hi. Firstly, thanks for atom-build!

I have recently stared using the left and right orientations and noticed an issue with the terminal window not filling the full height of the container. I have tried to fix this and also added a non persistent  cache for the width/height values so that they are maintained between switching orientation.

Thanks


C-Bouthoorn:
Fixed linter issues here: https://github.com/abartelsmeier/atom-build/pull/1

abartelsmeier:
hi, thanks for that, just accidental fragment of leftover code. Made a few mistakes here... shouldn't have cancelled the pull request... we live and we learn! It's reopened now, #387.

PR #386, opened 2016-04-07
added option to refresh targets on demand

tasnad:
Hi,
I need this for a build provider which depends on the currently selected file.
This is due to the fact, that I usually don't add each (latex) project path to atom, but one path with subprojects.
If this gets merged, I could also publish my small provider.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -185,7 +185,7 @@ export default {
   },
 
   selectActiveTarget() {
-    if (atom.config.get('build.refreshTargetsOnDemand')) {
+    if (atom.config.get('build.refreshOnShowTargetList')) {
       this.refreshTargets();
     }
     const p = require('./utils').activePath();
--- a/lib/build.js
+++ b/lib/build.js
@@ -186,7 +186,7 @@ export default {
 
   selectActiveTarget() {
     if (atom.config.get('build.refreshTargetsOnDemand')) {
-        this.refreshTargets();
+      this.refreshTargets();
     }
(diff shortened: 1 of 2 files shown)


noseglid:
Could you not simply dispatch the `build:refresh-targets` command from your build provider?

tasnad:
I'm not quite sure, but I don't think so. The problem is, that build does not call the provider if the file to display in the editor is changed, so the targets aren't updated.
Please correct me if I'm wrong, as I'm not really familiar with JS...

C-Bouthoorn:
_Irrelevant sidenote:_ Isn't it easier to make an `addTask` function in `config.js`, which increases the counter ? It looks unnecessary to change the order of all following tasks when adding a new one.

Is it worth changing?

noseglid:
@C-Bouthoorn The order is not arbitrary. I've tried to put options which relate to each other next to each other (such as status bar visibility and status bar priority). As does this PR. The orders only need to be changed if putting an option somewhere in the middle.

@tasnad I'm okay with merging this, but I'd like to change the name of the setting to `refreshOnShowTargetList`. The description is fine.

tasnad:
@noseglid You're right that might be the appropriate name. I updated the name. Thanks :)
@C-Bouthoorn Wouldn't it be possible to use a list instead of a map? Than the "order" field could be calculated and filled in somehow (depending on the order in the list)?

noseglid:
Cool, thanks!

tasnad:
Alright, thanks!

PR #387, opened 2016-04-07
Fix terminal height not filling container panel

abartelsmeier:
Fix for problem with terminal not filling container panel vertically when left or right orientation setting is selected.

Thanks for atom-build!


lib/build-view.js
```diff
@@ -131,12 +132,15 @@ export default class BuildView extends View {
   }
 
   resizeToNearestRow() {
+    if ( typeof this.cache === 'undefined' ) return;
```
noseglid: Is this ever the case? It's being assigned in the constructor

lib/build-view.js
```diff
@@ -158,10 +162,16 @@ export default class BuildView extends View {
     if (0 === w || 0 === h) {
       return;
     }
-
+    const orientation = atom.config.get('build.panelOrientation') || 'Bottom';
+    if ( orientation === 'Left' || orientation === 'Right' ) {
+      this.cache[orientation].width = $('.build .output').width();
+    } else if ( orientation === 'Top' || orientation === 'Bottom' ) {
+      this.cache[orientation].height = $('.build .output').height();
```
noseglid: Previously the height was taken from  `this.terminalEl`. This is not equivalent if `terminalEl` has margin, borders or padding. I think this should still grab the height from `this.terminalEl`, or calculations will off.

Was this change the thing that actually solved the issue?

lib/build-view.js
abartelsmeier: This confused me too... was getting an error because resizeToNearestRow was being called before cache was initialised in constructor. How?! Had a chance to look into this today and turns out, atom.config.observe('editor.fontSize') returns immediately with the current fontSize, which then calls resizeToNearestRow through fontSizeFromConfig. I've moved the cache init ahead of the config subscriptions, and removed the undefined check.

lib/build-view.js
abartelsmeier: I'm not too sure what the best approach is here. This did fix the issue I described but agree that there should be a more complete solution. I'll take a look at it again.

noseglid: Without looking any close, I remember there were some issues with the terminal element not "automatically expanding the height" - which I think is how any block DOM element behaves. Maybe [flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes) can be used?

abartelsmeier: Have committed a 'flex' approach, which seems to work, thanks! Still need to resize the terminal initially.

I've also slipped another fix in with this commit to make the build output window line up with the item-views when the panel is on the left or right. Basically the header is resized initially to be the same height as the tabs/padding above the editor window. Would you prefer that to be a separate pull request?

lib/build-view.js
```diff
@@ -190,9 +186,23 @@ export default class BuildView extends View {
...
+      this.resizeHeading();
+    }
+    this.fixTerminalElHeight();
     this.resizeToNearestRow();
   }
 
+  resizeHeading() {
+    const headingHeight = $('.item-views').position().top;
+    $('.build .heading').css('height', `${headingHeight}px`);
+  }
```
noseglid: I'm not sure using `.item-views' top position here works. What if there's another panel (e.g. `linter` above the build window?

noseglid: Also, why do we need to resize the heading?

noseglid:
It seems to be working fairly well. I asked a few questions. I wanna test this really thoroughly, because I find it quite tricky with a lot of room for mistakes. Bear with me.

lib/build-view.js
abartelsmeier: Its not strictly necessary, I just think it just looks neater with the themes I have tested with as it makes the build output panel line up with the editor view when the panel is left or right.
atom-material-ui
atom-dark-ui

abartelsmeier:
yep, no problem.

noseglid:
Did you consider what I asked about `.item-views` ?

abartelsmeier:
So this will make the build heading the same height as everything above item-views which (I think) is just tabs and padding? Im not sure if I understand this correctly; if there is a window above the build panel, will this affect the y pos of item-view`?

abartelsmeier:
If there is something above item-views then the heading would look strange I guess..

abartelsmeier:
heading resize removed, won't work for all cases.

noseglid:
I've tested this now and it works well! Thanks for your patience!

abartelsmeier:
you're welcome!

PR #393, opened 2016-04-12
Error matcher tweak

tasnad:
Hi, i'ts me again with my file-based target provider :)
I wanted to match errors in LaTeX targets, where there is the problem that the build logs don't contain the file name. My solution is to use the current file, if the match pattern did not contain a (?\<file\> RE ) named group.
If a target provider implements the file group, nothing changes. Only if it does not, build tries to get the current file and uses it as reference.
Maybe one could add an option for this, but in my opinion this is nothing the user should bother with.

Regards,
Tasnad

### DIFF
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -75,6 +75,8 @@ export default class ErrorMatcher extends EventEmitter {
         if (!match.file) {
           match.file = atom.workspace.getActivePaneItem().buffer.file.path;
         }
+      } catch {
+        console.log("build: Target provider did not give a 'file' and there is no valid open file!");
       } finally {
         // if no (real) file was open in the editor, match.file remains empty
         this.push(match);
--- a/lib/error-matcher.js
+++ b/lib/error-matcher.js
@@ -71,6 +71,10 @@ export default class ErrorMatcher extends EventEmitter {
     const self = this;
     const matchFunction = function (match, i, string, regex) {
       match.id = 'error-match-' + self.regex.indexOf(regex) + '-' + i;
+      if (!match.file) {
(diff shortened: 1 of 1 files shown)


noseglid:
Why is it obvious that the currently edited file is the one we want if no `file` exists in the error match?

tasnad:
I know, it's not. That's why I said that nothing changes if the (?\<file\> RE) named group is provided. (I just noticed that my previous comment here got Markdowned and the group name disappeared, sorry for that. I fixed it now).
The Latex log output is a huge mess, so if it's not the current file, it will anyhow point into the void.

So in the end:
1) Nothing changes for providers which are properly written including the (?\<file\> RE) Regex group
2) It works for Latex (and potentially other languages), if it is really the current file.
3) Otherwise, the output is rubbish anyways

Cheers,
Tasnad

noseglid:
So, the reservation I have towards this is that the next person will come and say "why is it the current file. That's not true in my setup. In my setup it is always the first file in the tab view", or "It's always `index.php`, no matter what the output says.

It will be hard, if not impossible to find a "good default file".

Maybe you can create a build provider instead and use the matching function (https://github.com/noseglid/atom-build/pull/398) once that is merged.

oli-obk:
@tasnad you know that you can tell latex to output `file:line` errors by setting `-file-line-error`, e.g.:

```
pdflatex -file-line-error test.tex
```

tasnad:
@oli-obk  Thanks!
That actually solves the problem. I updated my [provider](https://atom.io/packages/build-make-file).
Now this pull request might ultimately be considered closed :)

PR #395, opened 2016-04-14
Allow to install package without GIT

ivankravets:
If git is not installed in the system (and it is OK), user will receive
```
Installing “build@0.59.0” failed.Hide output…
Failed to install build because Git was not found.

The build package has module dependencies that cannot be installed without Git.

You need to install Git and add it to your path environment variable in order to install this package.

You can install Git by downloading, installing, and launching GitHub for Windows: https://windows.github.com

Run apm -v after installing Git to see what version has been detected.
```

Please release 0.59.1. Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "getmac": "^1.0.7",
     "js-yaml": "^3.4.6",
     "node-uuid": "^1.4.3",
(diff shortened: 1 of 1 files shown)


noseglid:
Thanks for this!

I don't think it's critical enough to warrant a patch release. I'll have it shipped in the next version which shouldn't be too far off anyway.

PR #398, opened 2016-04-18
allow execution of function to generate matches

oli-obk:


noseglid:
I like this idea. But I think it requires some thought before something like this is merged. All of a sudden `this.currentMatch` will be part of the API and the way matches are pushed onto it must be a locked API.

If you're still interested in doing this here's a few pointers to help you move forward:

  * I think the match function should return an array of matches rather than "pushing on `this`".
  * The object which is returned in above should only have keys which are described in the error matcher documentation (`file`, `col`, `col_end`, etc).
  * How to use a function as an error matcher must be documented in the README
  * Specs must be added

oli-obk:
Yea, this was simply a proof of concept impl

Returning a list sounds sensible, but I don't know much about javascript, it's all stackoverflowed together

Wrt only having the described fields, i don't see the point in checking.

I'll add docs once the design is finalized

oli-obk:
> I think the match function should return an array of matches rather than "pushing on this".

done

> The object which is returned in above should only have keys which are described in the error matcher documentation (file, col, col_end, etc).

This isn't true for regex now, and I don't see why it would be an issue to have more fields

> How to use a function as an error matcher must be documented in the README

done

> Specs must be added

I have no clue what `Specs` are in this context.

noseglid:
`Specs` are tests. See the `spec` folder. I suggest you create a new file, copy the topmost `beforeEach` and `afterEach` from another file to get the correct testing env, then write some specs to guarantee the functionality over coming releases

oli-obk:
I'm not sure what I'm doing wrong, a third of the specs are failing by default on my machine and the new specs I added are timing out. Is there a way to run just one spec file?

noseglid:
Yes, if you change `it` to `fit` it will *focus* that spec.

oli-obk:
thanks, that helped a lot. Specs are passing now.

spec/fixture/.atom-build.error-match-function.js
```diff
@@ -0,0 +1,22 @@
...
+  cmd: 'echo',
+  args: [ 'cake' ],
+  name: 'from js',
+  sh: true,
+  errorMatch: function (terminal_output) {
+    return [
+        {
+            file: '.atom-build.js',
+            line: '1',
+        },
```
noseglid: Could you also add `col` to one of the error match test cases?

noseglid:
Just one minor thing, then I'm super happy with this change!

oli-obk:
done

noseglid:
Thank you so much for this. Really great addition!

PR #400, opened 2016-04-19
Add selection replacement

### LINKED ISSUE #375: Allow customization based on current cursor selection
I often want to run a custom build target that uses the current selection to know what to run. For example,  I'd like to write a command that does `py.test {current-file} -k {current-selection}`. It runs the currently selected test in the current file.

evmorov:
Resolves #375


noseglid:
Very nice, thank you for this!

3noch:
@evmorov 👏  Thank you!

Issue #414, opened 2016-05-02
Increase terminal scrollback or allow configurable scrollback


yufanyufan:
The default scroll back of the terminal (1000 lines) is not enable to view the build log of a large project.

noseglid:
fixed with #415

PR #415, opened 2016-05-02
Fix Increase terminal scrollback or allow configurable scrollback #414

yufanyufan:
#414

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -109,4 +109,11 @@ export default {
     default: -1000,
     order: 15
   }
+  terminalScrollback: {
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -109,4 +109,11 @@ export default {
     default: -1000,
     order: 15
   }
+  terminalScrollback: {
```
noseglid: You missed a comma here.

yufanyufan: Fixed.

noseglid:
Thanks!

PR #416, opened 2016-05-04
normal font-weight in build output "terminal"

braver:
The 600 weight is fine for the panel header, but the output is much easier to read if it's at normal font-wight.

screen shot 2016-05-04 at 10 06 18

screen shot 2016-05-04 at 10 06 00


noseglid:
Thanks!

PR #418, opened 2016-05-08
enable traces (sub/child-errors) and custom "error types"

### LINKED ISSUE #408: Support warnings in addition to errors
Currently, the only supported property is "errorMatch". It would be great to have "warningMatch" and let linter show the warnings. Basically exactly the same behaviour as for errors except that they don't break the build ( by default ) and that they show up orange in linter :)

oli-obk:
This adds `warningMatch` and improves `functionMatch` to allow build providers to emit almost any kind of lint that is specified here: https://github.com/steelbrain/linter/wiki/Linter-API#messages

An example of a modified `build-cargo`:
pic of traces and custom error types

fixes #408

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -72,7 +72,7 @@ export default {
       atom.notifications.addError('Error matching failed!', { detail: message });
     });
     this.errorMatcher.on('matched', (match) => {
-      this.buildView.scrollTo(match[0]);
+      match[0] && this.buildView.scrollTo(match[0]);
     });
   },
 
--- a/lib/build.js
+++ b/lib/build.js
@@ -184,7 +184,7 @@ export default {
 
         let success = (0 === exitCode);
         if (atom.config.get('build.matchedErrorFailsBuild')) {
-          success = success && !this.errorMatcher.hasMatch();
+          success = success && !this.errorMatcher.getMatches().some(match => match.criticality === 'error' || match.type.toLowerCase() === 'error');
         }
 
         function extractRange(json) {
--- a/lib/build.js
+++ b/lib/build.js
@@ -187,15 +187,47 @@ export default {
           success = success && !this.errorMatcher.hasMatch();
         }
 
+        function extract_range(json) {
+          return [
+            [ (json.line || 1) - 1, (json.col || 1) - 1 ],
+            [ (json.line_end || json.line || 1) - 1, (json.col_end || json.col || 1) - 1 ]
+          ];
+        }
+        function normalize_path(path) {
+          return npath.isAbsolute(path) ? path : npath.join(cwd, path);
+        }
+        function type2severity(type) {
+          switch (type && type.toLowerCase()) {
+            case 'error': return 'error';
+            case 'warning': return 'warning';
+            case 'trace': return 'info';
+            case 'note': return 'info';
+            default: return null;
+          }
+        }
+        function clean_severity(severity) {
+          switch (severity && severity.toLowerCase()) {
+            case 'error': return 'error';
+            case 'warning': return 'warning';
+            case 'info': return 'info';
+            default: return null;
+          }
+        }
+
         const npath = require('path');
         this.linter && this.linter.setMessages(this.errorMatcher.getMatches().map(match => ({
-          type: 'Error',
+          type: match.type || 'Error',
           text: match.message || 'Error from build',
-          filePath: npath.isAbsolute(match.file) ? match.file : npath.join(cwd, match.file),
-          range: [
-            [ (match.line || 1) - 1, (match.col || 1) - 1 ],
-            [ (match.line_end || match.line || 1) - 1, (match.col_end || match.col || 1) - 1 ]
-          ]
+          filePath: normalize_path(match.file),
+          severity: clean_severity(match.severity) || type2severity(match.type) || 'error',
+          range: extract_range(match),
+          trace: match.trace && match.trace.map(trace => ({
+            type: trace.type || 'Trace',
+            text: trace.message || 'Trace in build',
+            filePath: trace.file && normalize_path(trace.file),
+            severity: clean_severity(trace.severity) || type2severity(trace.type) || 'info',
+            range: extract_range(trace),
+          })),
         })));
 
         if (atom.config.get('build.beepWhenDone')) {
--- a/lib/build.js
+++ b/lib/build.js
@@ -187,16 +187,16 @@ export default {
           success = success && !this.errorMatcher.hasMatch();
         }
 
-        function extract_range(json) {
+        function extractRange(json) {
           return [
             [ (json.line || 1) - 1, (json.col || 1) - 1 ],
             [ (json.line_end || json.line || 1) - 1, (json.col_end || json.col || 1) - 1 ]
           ];
         }
-        function normalize_path(path) {
-          return npath.isAbsolute(path) ? path : npath.join(cwd, path);
+        function normalizePath(p) {
+          return require('path').isAbsolute(p) ? p : require('path').join(cwd, p);
         }
-        function type2severity(type) {
+        function typeToSeverity(type) {
           switch (type && type.toLowerCase()) {
             case 'error': return 'error';
             case 'warning': return 'warning';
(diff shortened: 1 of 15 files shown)


lib/error-matcher.js
```diff
@@ -68,18 +68,37 @@ export default class ErrorMatcher extends EventEmitter {
...
-    this.regex.forEach((regex) => {
-      if (typeof regex === 'function') {
-        regex(this.output).forEach(function (match) {
-          this.push(match);
-        }.bind(this.currentMatch));
-      } else {
-        require('xregexp').forEach(this.output, regex, matchFunction.bind(this.currentMatch));
+
+    // first run all functional matches
+    this.functions && this.functions.forEach((f, functionIndex) => {
```
noseglid: since `this.functions` is set in the `set` function I feel we can safely assume this will always be an array

lib/build.js
```diff
@@ -180,22 +180,53 @@ export default {
...
-          type: 'Error',
+          type: match.type || 'Error',
           text: match.message || 'Error from build',
-          filePath: npath.isAbsolute(match.file) ? match.file : npath.join(cwd, match.file),
-          range: [
-            [ (match.line || 1) - 1, (match.col || 1) - 1 ],
-            [ (match.line_end || match.line || 1) - 1, (match.col_end || match.col || 1) - 1 ]
-          ]
+          filePath: normalizePath(match.file),
+          severity: cleanSeverity(match.severity) || typeToSeverity(match.type) || 'error',
```
noseglid: I feel this may be a bit too much. Why do we need to introduce both `type` and `severity` ?

lib/build.js
```diff
@@ -180,22 +180,53 @@ export default {
...
+        function normalizePath(p) {
+          return require('path').isAbsolute(p) ? p : require('path').join(cwd, p);
+        }
+        function typeToSeverity(type) {
+          switch (type && type.toLowerCase()) {
+            case 'error': return 'error';
+            case 'warning': return 'warning';
+            case 'trace': return 'info';
+            case 'note': return 'info';
+            default: return null;
```
noseglid: Hmm, I have concerns about this. where did `trace` and `note` come from ? It doesn't seem to have anything to do with the generic `build` tool. SInce linter supports `error`, `warning` and `info` I think we should keep those levels too.

But this comment would be unnecessary if we didn't have both `type` and `severity`.

noseglid:
This amount of code almost warrants moving all linter related code to a `linter-integration.js` file.

lib/error-matcher.js
oli-obk: javascript scares me xD I add checks everywhere. But yea, I'll remove this check

lib/build.js
oli-obk: the `Explanation` example uses `Explanation` as the `type` and `info` as the severity. Alternatively we can treat everything that's not `Error`, `Err`, `Warn` or `Warning` as `info`

lib/error-matcher.js
oli-obk: ah, no it's not always set, only if `functionMatch` is a value.

oli-obk:
I removed the criticality, and am parsing the type now. I added a few possible type values. Matches with unknown type are turned into `error`, Children of matches with unknown type are turned into `info`.

oli-obk:
> This amount of code almost warrants moving all linter related code to a linter-integration.js file.

I tried that, I failed, not sure what I did wrong.

oli-obk:
nevermind, I found it :) works now

noseglid:
Shaping up nicely! I'll give this a test and a final review later today!

lib/error-matcher.js
```diff
@@ -68,42 +68,61 @@ export default class ErrorMatcher extends EventEmitter {
...
+        if (typeof f !== 'function') {
+          this.emit('error', 'found functionMatch that is no function: ' + typeof f);
+          return false;
+        }
+        return true;
+      });
+    }
+    this.regex = {
+      Error: this._prepareRegex(target.errorMatch),
+      Warning: this._prepareRegex(target.warningMatch)
```
noseglid: Any reason why the keys here are capitalized?

noseglid:
Functionalitywise this is great, but I think we need some more documentation around this, with examples on how to add trace etc. The image you have in this PR would also be nice to spark interest!

lib/error-matcher.js
oli-obk: atom-build currently prints `Error` capitalized, I'm generating that name from the field name.

lib/error-matcher.js
```diff
@@ -68,42 +68,61 @@ export default class ErrorMatcher extends EventEmitter {
...
+        return match;
+      }));
+    });
+    // then for all match kinds
+    Object.keys(this.regex).forEach(kind => {
+      // run all matches
+      this.regex[kind] && this.regex[kind].forEach((regex, i) => {
+        regex && require('xregexp').forEach(this.output, regex, (match, matchIndex) => {
+          match.id = 'error-match-' + i + '-' + matchIndex;
+          match.type = kind;
```
oli-obk: here I use the field name as the type

oli-obk:
I'll think of an introductory example, but I'm not much of a javascript programmer, so my examples might be rather limited.

noseglid:
I think one of the specs should do it. Not much is required so one can get started.

lib/error-matcher.js
noseglid: Allright, makes sense!

oli-obk:
I added a new example and a spec testing the example

noseglid:
I'm totally happy with the code and docs now.
Something broke on Windows though :(

oli-obk:
probably the `/` path separators, I'll investigate

spec/build-error-match-spec.js
```diff
@@ -577,6 +611,68 @@ describe('Error Match', () => {
...
+  });
+
+  describe('when using function matches', () => {
+    it('should be possible to keep state from previous lines', () => {
+      expect(workspaceElement.querySelector('.build')).not.toExist();
+      fs.writeFileSync(directory + '.atom-build.js', fs.readFileSync(functionChangeDirs));
+      fs.writeFileSync(directory + 'change_dir_output.txt', fs.readFileSync(__dirname + '/fixture/change_dir_output.txt'));
+      fs.mkdirSync(directory + 'foo');
+      fs.mkdirSync(directory + 'foo/src');
+      fs.writeFileSync(directory + 'foo/src/testmake.c', 'lorem ipsum\naquarium laudanum\nbabaorum petibonum\nthe cake is a lie');
```
noseglid: THis could be the issue. Should probably use `path.join`.

oli-obk:
atom still keeps a handle on a file in the directory while we try to delete the directory after each spec run. This issue is talked about here: https://github.com/isaacs/rimraf/issues/72

Not sure how to fix this, because I have no idea how to close a tab in code. And even then, atom might keep the handle longer than it should.

Can we just stop deleting the specs after running them? They are in a temp directory with a unique id anyway.

noseglid:
But this only happend after your last change. Previously it was building fine!

oli-obk:
I'm actually creating subdirectories. Atom might somehow lock them.

noseglid:
The stack trace is originating from `https://github.com/noseglid/atom-build/pull/418/files#diff-bd51e8ec703453db2d933b31b969ab0aR58`.
I'd suggest putting a try-catch around that, if we fail to clean up, buhoo. It'll all work anyway and the next tests gets a new temporary directory anyway.

oli-obk:
> I'd suggest putting a try-catch around that, if we fail to clean up, buhoo. It'll all work anyway and the next tests gets a new temporary directory anyway.

That's what I meant, for simplicity we can just not clean up at all :) It's not like we're producing huge files during the specs.

noseglid:
I'd feel better with a try-catch than completely removing it though. When I do development I could potentially end up with a lot of temporary files which could've been cleaned up

noseglid:
Just put a comment in the catch
```
catch (err) {
  // Failed to clean up, ignore this.
}
```

run `npm test` locally to lint check it.

oli-obk:
I edited it throug the github interface b/c I wasn't on a pc that I had set up

noseglid:
Merged! Thanks so much for this and all your patience!

PR #437, opened 2016-06-02
remove ANSI escape sequences

patgithub:
remove ANSI escape sequences before matching any regular expression for easy parsing of formatted/coloured build messages.


noseglid:
This has been discussed multiple times and been decided against. If you want to do error matching you should not output any ansi escape sequences. Atom build does not advertise a pty, so most sane commands would not output any ANSI escape sequences.

PR #448, opened 2016-06-25
Convert platform-specific path delimiters to native platform delimiters.

Talon1024:
Use a more elegant method of passing `exec` and its arguments to child_process.spawn()

Also:
- Convert Windows path delimiters to Unix path delimiters on Unix platforms.
- Convert Unix path delimiters to Windows path delimiters on Windows platforms.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -130,6 +130,7 @@ export default {
     }).then(target => {
       const replace = require('./utils').replace;
       const pathFix = require('./utils').pathFix;
+      const cmdQuote = require('./utils').cmdQuote;
 
       const env = Object.assign({}, process.env, target.env);
       Object.keys(env).forEach(key => {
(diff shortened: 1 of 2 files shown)


noseglid:
Could you also add a test for this? It's been broken a few times already.

Talon1024:
I'm still working on this.

Talon1024:
I've decided to change this and just use a more elegant way of passing the `exec` and `args` to the child process.

Judging by the way you've wrote your unit tests, the `sh` option is only supposed to be true if the `cmd` string is a series of bash/shell commands, and not the name of a build script or build command.

noseglid:
Well.. if `sh` is true. The command will be spawned with a shell (e.g. `bash` or `cmd.exe`.) If it is false the OS will directly [exec](http://www.freebsd.org/cgi/man.cgi?query=exec&sektion=3) the command.

Exec has a limitation in that it doesn't support arguments "in the command". Other than that either is fine.

I tend to avoid `sh` because it feels faster to not start a shell and then spawn the command in there - but I don't think it has any noticeable impact on execution time in most cases.

noseglid:
Closing as inactive. Please open a new PR if you intend to keep working on this.

PR #449, opened 2016-06-27
Add more menu actions

Gawdl3y:
This adds a few more of the commands to the package menu, and changes the label slightly on the existing two.


noseglid:
Thanks! 👍

PR #450, opened 2016-06-27
Allow consumeBuilder to accept arrays

Gawdl3y:
This adds the capability for a single package to provide multiple build providers.

### DIFF
--- a/create-provider.md
+++ b/create-provider.md
@@ -19,7 +19,7 @@ The package should integrate via the service API. This is typically done in `pac
 ```
 
 The `build`-package will then call `providingFunction` when activated and expects an
-ES6 class in return. The next section describes in detail how that class is
+ES6 class or an array of classes in return. The next section describes in detail how that class is
 expected to operate.
 
 ## The provider implementation
--- a/lib/build.js
+++ b/lib/build.js
@@ -302,11 +302,11 @@ export default {
   },
 
   consumeBuilder(builder) {
-    this.tools.push(builder);
+    if (Array.isArray(builder)) this.tools.push(...builder); else this.tools.push(builder);
     this.targetManager.setTools(this.tools);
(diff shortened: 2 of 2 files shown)


noseglid:
Hmm.. do you have an example of an integration which provides multiple providers?
Wouldn't it make more sense to have separate builder providers instead?
There's no limit to the amount of build targets a provider may give.

Gawdl3y:
Mostly for reasonable separation of logic and cleanliness, and partially for #451.
Take my [build-papyrus](https://github.com/Gawdl3y/atom-build-papyrus) package as an example. Papyrus has multiple formats it can be built to: assembly (`.pas`), and a fully-compiled form (`.pex`). Currently, my package only supports full compilation, but I'd like it to support assembly/disassembly as well. I wouldn't want to shove the assembly/disassembly-related logic into the existing build provider, as they would work rather differently from each other. I'd end up with one big, convoluted mess of code.

noseglid:
I don't get it. Aren't you arguing against your point? Why not create a completely separate build probider for the assembly/disassembly functionality?

Gawdl3y:
build-papyrus encompasses all Papyrus-related building, not just script compilation. If I were to separate each component into separate packages, that'd be duplicating a ton of shared code between packages (config loading, game definitions, package settings, etc). It'd also be less user-friendly, which is an important aspect to me since there are loads of brand new scripters/devs coming into the scene that may not have any prior experience (I wrote and am continuously updating [a guide for using Atom](http://www.creationkit.com/fallout4/index.php?title=Atom) for Papyrus).

noseglid:
I see the appeal. I think it may be beneficial to publish an npm package for the shared code, but if you think it makes more sense to have it in the same atom plugin (that have multiple providers) I'm okay with merging this. It doesn't add much complexity.

PR #452, opened 2016-06-28
Send SIGINT rather than SIGTERM

noseglid:
While SIGTERM may be the "correct" choice from a documentation
standpoint, SIGINT is more widely used. Most applications are
designed to run in a terminal and most (all?) send SIGINT
when issuing ^C.

This PR simulates this behavior from atom-build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,7 @@ Option            | Required       | Description
 `warningMatch`    | *[optional]*   | Like `errorMatch`, but is reported as just a warning
 `functionMatch`   | *[optional]*   | A (list of) javascript functions that return a list of match objects
 `keymap`          | *[optional]*   | A keymap string as defined by [`Atom`](https://atom.io/docs/latest/behind-atom-keymaps-in-depth). Pressing this key combination will trigger the target. Examples: `ctrl-alt-k` or `cmd-U`.
+`killSignals`     | *[optional]*   | An array of signals. The signals will be sent, one after each time `Escape` is pressed until the process has been terminated. The default value is `SIGINT` -> `SIGTERM` -> `SIGKILL`. The only signal which is guaranteed to terminate the process is `SIGKILL` so it is recommended to include that in the list.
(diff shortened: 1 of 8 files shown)


oli-obk:
:+1: wonderful. This probably fixes the issue that sometimes the running application isn't terminated if a new build is triggered before the old one finishes

noseglid:
I hope so. Looks like with `SIGTERM`, the application should terminate as quickly as possible, and is allowed to leave some resources. Not good obviously. I think this will be a better approach

Gawdl3y:
Perhaps a target config setting could be added to continue using `SIGTERM`, in case it's ever needed for some reason?

noseglid:
@Gawdl3y perhaps if there ever arose a need for it. I highly doubt it though. To send `SIGTERM` in a terminal you'd have to invoke `kill` and find the pid. I'm willing to bet almost all programs respond "cleaner" to `SIGINT`.

noseglid:
What might make sense is to progress this way for each press of Escape: `SIGINT` -> `SIGTERM` -> `SIGKILL`. They're defined with increasing "severity" - where the last one `SIGKILL` will not even give the process a chance to intercept. The OS will just take back all resources it's holding.

oli-obk:
What would you do then in case the build is restarted before the previous one is finished?

noseglid:
You can't. It will not start a new process until the previous one has been terminated. Issuing `build` should follow the same progression I think. First time, send `SIGINT`, when child terminates run build again. Once more and it will send `SIGTERM` and wait. Finally it's a `SIGKILL` which guarantees that it terminates and the build can be restarted.

noseglid:
@oli-obk @Gawdl3y I'd love your input on this.

oli-obk:
I tried to grok the code. But my javascript-closure-foo is not strong enough.

The idea seems fine to me, even though it'll mean my regular use case will be ESC + ESC + F6.

noseglid:
That's a good point. For a lot of people I'd assume. Now I'm thinking I
want to make it configurable, and keep the default at `SIGTERM` ->
`SIGKILL`. Maybe in the build configuration even.

noseglid:
I made it configurable on the build command. I think that makes sense, since commands may have different needs, so a global configuration might not cut it.

oli-obk:
Awesome! Can it only be set from js or are the names strings?

Gawdl3y:
Your solution definitely gets a :+1: from me.

noseglid:
Works with any format. Build providers can also set it

noseglid:
Thanks for help and suggestion @oli-obk and @Gawdl3y

PR #455, opened 2016-07-08
Linter's html field support

### LINKED ISSUE #428: html error text
Since `linter` supports html error text, would it make sense to add a `html_message` field/regex match ?

Note: my actual use case is markdown, but converting that to html is trivial.

alygin:
Added support for the Linter's message `html` field via `html_message`.

fixes #428

### DIFF
--- a/lib/linter-integration.js
+++ b/lib/linter-integration.js
@@ -32,14 +32,14 @@ class Linter {
     this.linter.setMessages(messages.map(match => ({
       type: match.type || 'Error',
       text: !match.message && !match.html_message ? 'Error from build' : match.message,
-      html: match.html_message,
+      html: match.message ? undefined : match.html_message,
       filePath: normalizePath(match.file),
       severity: typeToSeverity(match.type),
       range: extractRange(match),
       trace: match.trace && match.trace.map(trace => ({
         type: trace.type || 'Trace',
         text: !trace.message && !trace.html_message ? 'Trace in build' : trace.message,
-        html: trace.html_message,
+        html: trace.message ? undefined : trace.html_message,
         filePath: trace.file && normalizePath(trace.file),
         severity: typeToSeverity(trace.type) || 'info',
         range: extractRange(trace)
--- a/lib/linter-integration.js
+++ b/lib/linter-integration.js
@@ -31,7 +31,8 @@ class Linter {
     }
     this.linter.setMessages(messages.map(match => ({
       type: match.type || 'Error',
-      text: match.message || 'Error from build',
+      text: !match.message && !match.html_message ? 'Error from build' : match.message,
+      html: match.html_message,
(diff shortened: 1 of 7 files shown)


alygin:
@noseglid, could you, please, help me with this PR? It seems like the build and all tests are good but Travis is stalling for no obvious (at least for me) reason. Have you ever encountered such a problem in this project before? Or did I break something?

lib/linter-integration.js
```diff
@@ -31,13 +31,15 @@ class Linter {
     }
     this.linter.setMessages(messages.map(match => ({
       type: match.type || 'Error',
-      text: match.message || 'Error from build',
+      text: !match.message && !match.html_message ? 'Error from build' : match.message,
+      html: match.html_message,
```
noseglid: If `message` *and* `html_message` is set, it will set both to the linter [which is an error](https://github.com/steelbrain/linter/wiki/Linter-API#messages).

noseglid:
I think it just stalled a little bit on travis's end. I restarted it and it passed. Sometimes weird stuff happens :)

lib/linter-integration.js
alygin: Yes, you're right. Here I saw two options: either we check it on our side and throw an error or let Linter make all the checks and throws errors. Now, other checks are made by Linter (for instance, that `path` is non-empty) so I decided to leave it this way. Although, if you think it must be the builder's responsibility to check the duplication, I'll add it. I think both attitudes have pros and cons and I have no strict preference other than being consistent in this question.

alygin: And, of course, there's a third option - just ignore either `message` or `html_message` when both are set. It will reduce quantity of errors seen by the end user, but it will also hide problems in the provider code. Maybe it's better to show an error and let the provider developer know that he or she did something wrong.

Another alternative would be to ignore `message` if it equals to `html_message` and only use the latter. I think `html_message` must be preferred because in case of html-tags absence it behaves the same way as `message`.

What do you think?

noseglid: I don't think we should invoke the linter API incorrectly. Since `message` is the most used I think it should have precedence. So if both are set, don't send `html_message` to linter. Does that sound okay to you?

alygin: Sure, that sounds reasonable as long as it is documented. I'll make the necessary changes.

noseglid: Awesome 👍

alygin:
Done with making `message` prioritized over `html_message`. Documentation and new tests are provided as well.

noseglid:
Sorry for taking so long to finish this one off. Summer, y'know :)

This is a great PR. Thanks for the work you've put into this!

alygin:
No worries at all. And thanks for merging! I hope build providers will make good use of the feature and come up with some cool stuff.

Issue #463, opened 2016-08-16
Support non-file specific errors


ProPuke:
Currently the filePath is mandatory when handling build errors.
It would be nice if it could be omitted so that general build errors could be displayed, too.

Some compilers or build systems may throw errors which are not specific to particular file locations, but should nonetheless be reported to users. While these can be seen if the build panel is open, if it is hidden these can go undetected.

It is possible to pass errors to Linter without a filePath (in which case it adds to the error total at the bottom, but does not display the error message anywhere). This could be used along with a standard atom error notification to show such errors.

Alternatively, Build could allow the provider to provide an `onErrors` callback for handling all reported errors including ones without filePaths, in which case it could perform special handling if it needed to. Or even just a general callback like `onData` so that it could manually handle any special output produced by the build command.

Cheers

C-Bouthoorn:
*I think* that you shouldn't rely on build to give you all information before you commit. Always make sure that it runs good in your terminal too, and maybe add CI (like Travis) to your project if it matters (production environment).

*I myself* would be frustrated if my error counter would give `2` and I could only find 1 error in my file, and no other error elsewhere.

*I also think* that it would be really hard for a package like `build` to distinguish between compiler-error and compile-error. (Especially since you don't provide an example)

*However*, I think there is / should be a function in the `build` API which allows a 3rd party to handle all errors? In that case, you could make an extension which handles these compiler-errors (which I myself haven't ever found yet)

ProPuke:
An example might be compiling C/C++
Sometimes you'll get compiler errors:
```
file.cpp:1:1: error: foo is foo
```
And sometimes you'll get linker errors:
```
Undefined reference to "foo"
```

They're both build errors, but atm there's no way of handling the latter with Build, even from custom Build packages. If you want to the only route currently is to drop Build and write your own package from scratch.

For now I'd suggest just using standard atom error notifications if there isn't a file location, so it isn't _just_ a counter with no informative message. (although if there were too many I realise this could flood the screen)

If not, just the ability to hook these errors or build output and handle it yourself would be fine. Then you could handle it appropriately from your own package.

C-Bouthoorn:
In reply to your C/C++ linking error; as far as I know, g++ (and possibly gcc) gives the file with the error even on linking errors.


```cpp
// test.cpp 
#include "linkerror.h"

int main() {
  Link link;
  link.error();
}
```

```cpp
// linkerror.h 
class Link {
public:
 void error();
};
```

```
[...]

PR #465, opened 2016-08-16
added: non-file specific errors

ProPuke:
`errorMatch` and `warningMatch` may now catching a the `file`.

These errors will be grouped together, and displayed via an atom error notification popup (as Linter does not currently support displaying errors without file locations).

see #463

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ the message text by setting `html_message` instead of `message`. If both
 
 Option            | Required       | Description
 ------------------|----------------|-----------------------
-`cmd`             | *[required]* | The executable command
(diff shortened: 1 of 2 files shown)


C-Bouthoorn:
Fixed linter errors in https://github.com/ProPuke/atom-build/pull/1

noseglid:
Joined with the answer in #463, I think this may not be the right way to go. I'm not comfortable with removing the requirement on `filePath`, and introducing yet another way to display errors will cause more confusion than it will actually help people I think.

Thanks for the proposal, but I must decline.

PR #477, opened 2016-10-02
Allow copying text from build output

AndiDog:
Copying text out of the build output div isn't working right now (tested with Atom 1.10.2 on Mac OS X). Since no `user-select` override is used, I guess it must be because of the way `core:copy` is mapped to keys by Atom's default keymaps – Cmd+C doesn't trigger `core:copy` when selecting text in the output console. This change enables copying text.

### DIFF
--- a/keymaps/build.json
+++ b/keymaps/build.json
@@ -13,6 +13,12 @@
     "ctrl-alt-h": "build:error-match-first",
     "ctrl-alt-t": "build:select-active-target"
   },
+  "body.platform-darwin .build .output": {
+    "cmd-c": "core:copy"
+  },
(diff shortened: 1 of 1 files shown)


noseglid:
It's working well for me. (also Mac, and Atom 1.10.2).
copy

Could there be something else interfering?

AndiDog:
Interesting. For me, neither `native!` nor `core:copy` were triggered. I'll check back with all other packages disabled to find out why it doesn't work for me.

AndiDog:
I cannot reproduce it anymore, so either it was resolved by an update of Atom, or just magically working. Nevermind - thank you!

noseglid:
👍

PR #501, opened 2017-02-21
Fix issue with incorrect path for text editor

ivankravets:
Atom >= 1.13 doesn't return a real path for `textEditor.getPath()`. In this case, affected projects will not be available for "build".

For example, project is located in `/private/tmp/myproject`
* `fs.realpathSync('/private/tmp/myproject') === '/private/tmp/myproject'`
* `textEditor.getPath() === '/tmp/myproject'`
* `fs.realpathSync(textEditor.getPath()) === '/private/tmp/myproject'`

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -32,7 +32,7 @@ const activePath = () => {
   return atom.project.getPaths().sort((a, b) => (b.length - a.length)).find(p => {
     try {
       const realpath = fs.realpathSync(p);
-      return textEditor.getPath().substr(0, realpath.length) === realpath;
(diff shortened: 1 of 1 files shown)


oli-obk:
Hmm. This fails some oft our Tests... Maybe they need that change, too?

ivankravets:
@oli-obk that is not a problem of this PR. See [previous commits](https://github.com/noseglid/atom-build/commits/master)  that also are failed.

oli-obk:
Oh, I'll check it out

noseglid:
I am sorry for being terribly slow on this.

The PR looks good. Would you mind rebasing on latest `master` where I've fixed the specs?

Issue #447, opened 2016-06-19
Freeze when "Select Active Targets"


xwvvvvwx:
# Description

Atom freezes whenever I press `cmd-alt-t` or select `Select Active Targets` in the command palette.

I am on Atom 1.8.0 (OS X 10.11.5), and `Build` is the only enabled plugin.

I am using the Genesis theme (https://atom.io/themes/genesis-ui).

The Genesis theme works fine with all my other plugins. 
(if it is a problem in the theme, please let me know and I will file a bug in the appropriate project)
# Steps
1. Launch atom
2. Press cmd-alt-t
# Outcome
1. Atom is frozen
# Recording

![:(](https://media.giphy.com/media/xT0GqkLacmPQZMVdIc/giphy.gif)

noseglid:
Do you get a crash report or anything actionable?
I'm not sure what to do with this report. It's not crashing for me, I have the exact same setup.

davidterry01:
@noseglid feeling pretty silly now, am also no longer able to reproduce the issue. 

Closed, sorry for taking up your time :/

ndugger:
I'm actually experiencing this issue right now.

My atom is up to day, the build package is up to date... I'm not sure what's happening.

icidasset:
I have the same issue, no idea where it's coming from and no crash reports (haven't waited longer than 5 minutes though).

noseglid:
When did it start to happen? What have changed? Does a restart solve it? Reinstall?

icidasset:
@noseglid I just installed it, along with the `build-stack` package. Restarting atom and reinstalling the package doesn't fix it. Also tried to disable the `build-stack` package, no changes. Tried to use the dev-tools to find the problem, but that freezes too. Anything else I could try?

noseglid:
I'm guessing somethings ends up in an infinite loop. That's normally what happens when everything freezes. Is (at least) 1 CPU running at 100% when it's frozen?

If so, this could be a way to debug further: https://www.joyent.com/blog/mdb-and-node-js. I have never used it for nodejs, unfortunately but something like that is the only way forward I can think of at least.

icidasset:
Yup.

<img width="527" alt="screen shot 2016-06-21 at 12 41 42 pm" src="https://cloud.githubusercontent.com/assets/296665/16226853/a1aef448-37ad-11e6-8071-a97fb7cf8da7.png">

I will try to look into that mdb thing.

noseglid:
[...]

PR #502, opened 2017-02-27
Fix #484, #447

### LINKED ISSUE #484: Infinite loop if no target
I got the infinite loop of calls `isEligible` method for all build-providers. It happens if I have no any `.atom-build.*`-file;

For localise & reproduce this issue I was create a new atom-plugin with empty provider contains `isEligible() { console.debug("LOOP HERE"); return false; }`

atom-build loop issue mov opt

Steps for reproduce:
1. Open any/new project;
2. Call `build:select-active-target` (press F7 by default)

sphaerophoria:
In TargetManager, getTargets, and by extension fillTargets, has an
implicit requirement that pathTarget.targets is not empty. If it is we
re-call refreshTargets which would end up with infinite recursion. We
don't actually need to call fillTargets as we already know which targets
are available, we can just set them ourselves

### DIFF
--- a/lib/target-manager.js
+++ b/lib/target-manager.js
@@ -122,6 +122,8 @@ class TargetManager extends EventEmitter {
 
         pathTarget.targets = settings;
         pathTarget.loading = false;
+
+        return pathTarget
       }).catch(err => {
         atom.notifications.addError('Ooops. Something went wrong.', {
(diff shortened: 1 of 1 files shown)


sphaerophoria:
I seem to have missed some of the side-effects of this change closing for the time being.

PR #503, opened 2017-02-27
Fix #484, #447

### LINKED ISSUE #484: Infinite loop if no target

sphaerophoria:
In TargetManager, getTargets, and by extension fillTargets, has an
implicit requirement that pathTarget.targets is not empty. If it is we
re-call refreshTargets which would end up with infinite recursion. We
have enough information to retrace the steps fillTargets currently takes so
we can just set them ourselves.

There are three failing tests with my current implementation, however these three tests were failing before I made any changes.

### DIFF
--- a/lib/target-manager.js
+++ b/lib/target-manager.js
@@ -122,6 +122,8 @@ class TargetManager extends EventEmitter {
 
         pathTarget.targets = settings;
         pathTarget.loading = false;
+
+        return pathTarget;
       }).catch(err => {
         atom.notifications.addError('Ooops. Something went wrong.', {
           detail: err.message,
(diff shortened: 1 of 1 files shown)


noseglid:
I'm sorry for being so slow to reply to this.
I think this looks good initially. I wanna test it a bit more.
What would really help is if we could add a spec, so we can guarantee it doesn't reappear.

noseglid:
Commited this in #520

PR #511, opened 2017-04-20
Pass a replacements object to preBuild

### LINKED ISSUE #510: Pass resolved replacements to preBuild
I want to selectively apply some `{SELECTION}` replacements in my `preBuild` function but I can't find a way to access those resolved strings. Would it be possible to pass those as arguments into `preBuild`? Perhaps the function signature could be changed to `preBuild(replacements)` where `replacements` is an object with properties `file_active`, etc that just contain the strings that these replacements would resolve to.

0xdec:
Fixes #510.
No tests yet as I don't know the best way of testing this.


noseglid:
As discussed in #510

PR #512, opened 2017-04-20
Fix spelling: succeding -> succeeding

0xdec:
I searched through several build providers to see if this would break anything, and I haven't yet found any uses of this hook, but this needs more extensive searching to be sure.

### DIFF
--- a/spec/build-hooks-spec.js
+++ b/spec/build-hooks-spec.js
@@ -7,7 +7,7 @@ import specHelpers from 'atom-build-spec-helpers';
 describe('Hooks', () => {
   let directory = null;
   let workspaceElement = null;
-  const succedingCommandName = 'build:hook-test:succeding';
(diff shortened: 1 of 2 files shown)


noseglid:
I think @ivankravets is using it. Am I wrong?

ivankravets:
We use preBuild/postBuild hooks. See https://github.com/platformio/platformio-atom-ide/blob/develop/lib/services/build.js#L258

0xdec:
I just searched through every repository in the list of [build providers](https://atombuild.github.io/) and couldn't find anyone using it. The only place I found it on GitHub was in many users atom configuration repos (dotfiles), which shouldn't matter.

noseglid:
This just changes the specs and an obvious spelling error.
Would you mind rebasing on latest `master` where I've fixed the specs?

noseglid:
Thanks!

PR #542, opened 2017-10-02
fix refresh targets for json files

### LINKED ISSUE #527: Refresh targets ignores .atom-build.json
When I add a new target to my .atom-build.json and then do "Packages -> Build -> refresh targets" the new target does not appear in the list of available targets. I have to restart atom to have those changes reflected. Same for removing a target or updating its cmd

wiggisser:
seems require(filename) does not reread a changed json  file
fixes #527

### DIFF
--- a/lib/atom-build.js
+++ b/lib/atom-build.js
@@ -8,8 +8,8 @@ function getConfig(file) {
   delete require.cache[realFile];
   switch (require('path').extname(file)) {
     case '.json':
-       return JSON.parse(fs.readFileSync(realFile));
-       
+      return JSON.parse(fs.readFileSync(realFile));
+
     case '.js':
       return require(realFile);
 


noseglid:
I guess the issue would still remain for a `js` file though.

wiggisser:
I set a breakpoint directly before deleting from require.cache. At this breakpoint I see the following

1) When I load a project, `require.cache[realFile]` is undefined (as expected)
2) When I change the .atom-build.json file `require.cache[realFile]` contains the old configuration (as expected)
3) When I change the .atom-build.json file again `require.cache[realFile]` is undefined (not as expected)

It seems, that after the first time, the config is deleted from the require.cache, it is never added again, and all calls to require return the initial configuration, like it was when the file was read for the first time.

noseglid:
Did you do these tests with or without the fix this PR suggests?

wiggisser:
Without the PR, ie I did a `require(realFile)` also for json files

wiggisser:
I did a bit more of debugging and tried to step into the `require` call. Seems that atom/electron has some custom implementation of the require call and a separate module cache which seems not to be affected by `delete require.cache[...]`. Thus, require is always returning the module from that custom module cache

wiggisser:
I filed an issue with atom (https://github.com/atom/atom/issues/16435) regarding this, they closed it with "won't fix", so I guess using plain `require` won't work for updating the build configuration.

I think I found a more general solution, but as I already deleted my fork of this repo I can't update the pull request, so I'm clsoing it and creating a new one ...

noseglid:
That's unfortunate, but at least we know why (and why I have to restart Atom for every package upgrade!).

PR #546, opened 2017-11-16
close stdin in spawned child processes

doudou:
If we don't close stdin, subprocesses that would ask for user input
block forever. Since atom-build has no way to interact with the spawned
subprocesses, just close it to have them fail.


noseglid:
Yea, that makes sense.

PR #568, opened 2018-08-20
Lint -> Atom IDE Diagnostics

dragonzap:
Replace old Linter to default Diagnostics tool.


fargies:
👍 makes this plugin compatible with atom-ide-ui diagnostics part (that only supports linter v2 messages).

